PR #21, opened 2022-10-25
specs/: fix local cross-references

t-lo:
Document cross-references do not work on the website if a `.md` file is referenced.

### DIFF
--- a/specs/boot_loader_specification.md
+++ b/specs/boot_loader_specification.md
@@ -739,7 +739,7 @@ There are a couple of items that are out of focus for this specification:
 
 [GUID Partition Table](https://en.wikipedia.org/wiki/GUID_Partition_Table)<br>
 [Boot Loader Interface](https://systemd.io/BOOT_LOADER_INTERFACE)<br>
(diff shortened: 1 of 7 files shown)


brauner (approved):

bluca:
This breaks following links on the github markdown visualizer, try to click on a link from https://github.com/uapi-group/specifications/blob/9580e78eb0bdf5232fa57b41be4d62265920c452/specs/discoverable_disk_image.md

t-lo:
> This breaks following links on the github markdown visualizer, try to click on a link from https://github.com/uapi-group/specifications/blob/9580e78eb0bdf5232fa57b41be4d62265920c452/specs/discoverable_disk_image.md

That's correct, the links are currently written for processing with Hugo. We'll need to find a way to make it work with both website and Github markdown in the future; I don't think we're the only ones running into this issue.

bluca:
The most important thing is that they work in github, as that's the most common way of reading things, so if there's no way to make them work in both, leave it broken in the website for now

t-lo:
@bluca I found a better way to make the links work, should now work for both rendered website and github.

bluca (approved):

PR #28, opened 2022-10-25
README: note derivation from XDG dir spec

bluca:


t-lo:
Nice work Luca!
Do you think it's worth adding a similar remark to the actual spec (in case someone deep-links it)? And maybe express intent to upstream the extensions if / where applicable?

bluca:
let's start from the readme, if more clarification is needed it can always be added

bluca:
upstream PR: https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/59

PR #51, opened 2023-02-06
rephrase boot_loader_specification.md

iviar:
somehow the same, I think said more clear


iviar:
While reading I started not liking the first "of" (in ... of each of ..., thinking maybe a "for each of" would avoid intending the attribute cumulative), then I rephrase

Foxboron (approved):

PR #53, opened 2023-03-21
version spec: deal with separator character, compare with other formats

bluca:


specs/version_format_specification.md
```diff
@@ -0,0 +1,85 @@
...
+5. If the remaining part of one of strings starts with `-`:
+   if the other remaining part does not start with `-`,
+   the string with `-` compares lower. Otherwise, both minus characters are skipped.
+6. If the remaining part of one of strings starts with `^`:
+   if the other remaining part does not start with `^`,
+   the string with `^` compares higher. Otherwise, both caret characters are skipped.
+6. If the remaining part of one of strings starts with `.`:
+   if the other remaining part does not start with `.`,
+   the string with `.` compares lower. Otherwise, both dot characters are skipped.
+7. If either of the remaining parts starts with a digit, numerical prefixes are
```
keszybz, line 51: Other points have:
> If condition: blah blah.

Maybe change this one to "If either of the remaining parts starts with a digit: ".

keszybz (changes requested):

specs/version_format_specification.md
```diff
@@ -20,6 +20,7 @@ A version string must use the following character set:
 - Minus (`-`) separates the version and release parts.
 - Tilde (`~`) is a prefix that always compares lower.
 - Caret (`^`) is a prefix that always compares higher.
+- Underscore (`_`) separates the version from the name of an artifact, if both are included in the same string.
```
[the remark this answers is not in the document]
bluca: I am aware, but the problem is that this is already factually the case for DDIs that we support, so we can't go back on that. We use `-` for prefixes within the names, and `_` to separate the name with the version. I know that this follows the Debian style rather than the RPM style, but I think that ship has sailed? This is not meant to impose a new format on packages (at least until we release systemd-packaged ;-) ), but just for the stuff that we do here (IE: DDIs, UKIs) so hopefully it should be ok?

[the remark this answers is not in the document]
bluca: @keszybz I could explicitly rename/reword this to include DDI/UKI in the title to ensure we don't get confused with packaging rules if that helps?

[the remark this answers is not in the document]
poettering: instead of declaring "_" a separator I'd just say "other characters, including '_', ':', ';', ',', '+' are not considered part of version strings compatible with this specification, and may hence can be used as separators preceeding or following a version string.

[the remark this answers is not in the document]
bluca: But again the problem is that we already use `_` in existing, released code for DDIs, and the purpose of this is to define how the version applies to DDIs, so I don't see how to get away without mentioning this explicitly?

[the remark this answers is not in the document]
keszybz: > instead of declaring `_"` a separator I'd just say "other characters, including `_`, `:`, `;`, `,`, `+` are not considered part of version strings compatible with this specification, and may hence can be used as separators preceeding or following a version string.

Sorry, but I don't think this works at all. As I wrote above, `_` is used in significant majority of strings that we would compare using this logic when using stuff that is packaged with rpm. If we try to disallow `_`, I think we'd not only break compatibility, but we're also unlikely to ever get any kind of compliance with the spec again, because suddenly almost everything would be in violation. E.g. every `bootctl list` and so on on Fedora.

The other characters are less problematic, because they are much less widely used.

As discussed before, `+` is used in semver. And semver is steadily gaining popularity. Even if we can't get rid of `+` used for boot counting, we shouldn't at least make the situation worse.

Similarly, `:` is used to seperate the epoch in full rpm version strings, but the epoch is not written to file names, so most of the time we wouldn't see it in comparisons. Right now we handle `:` in a reasonable way, as a separator. If both strings have an epoch, then it is compared numerically, and the right result occurs. rpmdev-vercmp has special support for skipping the epoch (`'0:253' == '253'`), which we obviously don't implement, but that's OK.

AFAIK, `;` and `,` are not used by anything. `;` is special for the shell, and `,` also is in some cases. If EFI firmwares support it, I think it would be reasonable to start using either of those two as the boot count separator.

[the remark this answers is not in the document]
poettering: > rpm, dpkg and semver are already not compatible:

yep, precisely. All I am saying it would be supernice to have this summary somewhere in the spec, to clarify things. in particular incompatibilities between the four specs would make sense to mention, as well as the "common core" of all four specs, and finally the "compressed charsets" of each spec.

but this is just wishful thinking on my side, nice stuff to have, to position this spec sensibly in relationship to the other specs, that's all. would be great to have as part of the "creation" of this new spec, but I totally can understand if you don't want to bother with this now (though in a way your comment here already covers 80% of what I am looking for, and hence cut just be copy/pasted into the spec), or ever.

specs/version_format_specification.md
```diff
@@ -0,0 +1,87 @@
...
+SPDX-License-Identifier: CC-BY-4.0
+---
+
+# Version Format Specification
+Many of the formats and concepts defined in these specifications rely on having
+a sort order for strings that include version components, and use it for various
+purposes, such as choosing the default boot entry in the
+[Boot Loader Specification](boot_loader_specification.md). This specification
+defines what format these version strings must follow, and how to
+deterministically compare them.
```
keszybz, line 14: This is still not ideal, because it suggests that the comparison is defined for the version component, when in fact it's defined for arbitrary strings. I can compare `systemd-253-6.fc38.i686` with `systemd-253-6.fc38.i686~` and I get a reasonable and deterministic result. This is a very important aspect of this algorithm and how it's used in practice.

specs/version_format_specification.md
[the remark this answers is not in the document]
keszybz: Yes, a section about incompatibilities would be nice-to-have.

specs/version_format_specification.md
bluca: dropped `version`

specs/version_format_specification.md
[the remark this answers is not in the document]
keszybz: > but then in the DDI and BLS specs pick a specific one and forbid the other

Pfffff. That's just terrible. DDI is new, can't we just adjust DDI version comparison to behave as BLS which has been around for 8 years?

specs/version_format_specification.md
keszybz: Maybe separate the second commit to a second PR? I'd like to get the first part merged and start adjusting URLs in other places and such…

keszybz:
The issues in the second commit are not addressed :(

specs/version_format_specification.md
```diff
@@ -63,6 +63,22 @@ strings are found to compare as different. In a loop:
...
+
+- [RPM Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/)
+- [Debian Policy](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version)
+- [Semantic Versioning](https://semver.org/)
+
+All of these, including the present document, share some commonalities but are also
+incompatible in some ways, as they all evolved in different environments. The main
+differences are as follows.
+
+- to separate components RPM uses `-`, DEB uses `_` and SemVer does not specify anything
```
keszybz, line 77: Hmm, semver only talks about the version part of the string… So by definition it doesn't care about how the version is separated from other components in a file name or such.

Also, RPM uses `-` with some positional logic. (I think the last two dashes separate the release and version, but I'd need to check that.)

specs/version_format_specification.md
```diff
@@ -63,6 +63,22 @@ strings are found to compare as different. In a loop:
...
+- to identify a pre-release suffix RPM and DEB use `~` and SemVer uses `-`
+- to identify a rebuild suffix DEB uses `+`, SemVer uses `.`, and RPM does not specify anything
```
keszybz, line 79: "does not specify anything" → "just increases the release number" ?

bluca:
> The issues in the second commit are not addressed :(

yes, still working on that, wanted to push the first change with the wording in the intro

specs/version_format_specification.md
bluca: updated

specs/version_format_specification.md
bluca: updated

specs/version_format_specification.md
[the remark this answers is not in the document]
keszybz: That's a bit of a bummer. I hope we can figure out a way to modify the DDI naming in a way that maintains some sort of backwards-compatibility but reconciles the name-version handling with the BLS and its version spec.

Or if that can't be done, just say in the DDI spec how the splitting should be done, and than that the version part is compared using the version spec.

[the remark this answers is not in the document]
poettering: @bluca which location specifically do you have in mind where we already rely on "_" as separator? portabled? or anywhere else?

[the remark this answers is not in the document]
bluca: portabled and sysupdated at least, not sure if it's wired in dissect itself yet https://www.freedesktop.org/software/systemd/man/sysupdate.d.html

poettering:
i wonder if we should define as part of the spec how to precisel map debian-style, rpm-style and semver-style version strings into this spec and back. And then implement "systemd-analyze version-convert" or so which converts these styles.

That way we could people clear help on how to map these things.

I mean, whatever we do, the incompatibilities of semver, rpm and dpkg version are nothing we can solve. We can just try to make our approach a bit friendly to handling these incompaitiblities in some fashion.

specs/version_format_specification.md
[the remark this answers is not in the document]
bluca: Update to avoid mention the topic and leave it unspecified, and changed the individual specs to specify the separator instead, would this work?

bluca:
> i wonder if we should define as part of the spec how to precisel map debian-style, rpm-style and semver-style version strings into this spec and back. And then implement "systemd-analyze version-convert" or so which converts these styles.
> 
> That way we could people clear help on how to map these things.
> 
> I mean, whatever we do, the incompatibilities of semver, rpm and dpkg version are nothing we can solve. We can just try to make our approach a bit friendly to handling these incompaitiblities in some fashion.

Would be nice, but I'd rather leave it for a different PR, as there's quite a bit to go through in this one already

specs/version_format_specification.md
```diff
@@ -62,6 +62,13 @@ strings are found to compare as different. In a loop:
    of the whole string), if the other substring has remaining letters, it compares higher.
    Otherwise, the comparison continues at the following characters at point 1.
 
+## Separating the Version from Other String Tokens
+The version is usually part of a larger string, usually containing other tokens such as an
+artifact's name. This specification does not mandate a specific separator between tokens,
+but it suggests to use a character that is not part of the allowed character set that forms
```
keszybz, line 68: This sentence is severely misleading. `_` **is** a part of the allowed character set in the version.

I don't think approaching this problem from this side is useful. It's not just a matter of changing a word or two. The basic fact is that there is no unambiguous separator in the common set of characters (`[a-zA-Z0-9._^~+-]`).

specs/discoverable_disk_image.md
```diff
@@ -18,3 +18,8 @@ The images use the GPT partition table verbatim, so it will not be redefined her
 a standard Linux filesystem (e.g.: squashfs), so again this will not be redefined here.
 The ([DPS](discoverable_partitions_specification.md)) defines the GUIDs to use and the format of the
 dm-verity signature partition's JSON content.
+
+## Image Version
+When versioning a DDI is appropriate, it is required to use the version format described in the
+[Version Format Specification](version_format_specification.md). As a separator between the version and
+other tokens in a string, the `_` character must be used.
```
keszybz, line 25: """
If the DDI is versioned, the version format described in the [Version Format Specification](version_format_specification.md) must be used.
The underscore character (`_`) must  be used to separate the version from the name (?) of the image.
"""

bluca: updated

specs/version_format_specification.md
keszybz: Yes, this is slopilly worded. When you check the actual algorithm below, it effectively handles `_` as a separator, similarly to `+` and other non-alphanumerical characters. We should update this introductory paragraph.

keszybz: No, they are not ignored. They separate parts of the version string. E.g. `a_b < ab`.

bluca: > No, they are not ignored. They separate parts of the version string. E.g. `a_b < ab`.

mmh but which part of the spec defines that?

keszybz: This is an iterative algorithm. In step 1, we skip leading characters. We then go to step 2, etc. After iterating though all listed characters classes, we return to step 1. So characters outside of the specified set are all non-sorting separators.

This should be worded better.

specs/version_format_specification.md
```diff
@@ -21,6 +21,8 @@ A version string must use the following character set:
 - Minus (`-`) separates the version and release parts.
 - Tilde (`~`) is a prefix that always compares lower.
 - Caret (`^`) is a prefix that always compares higher.
+- Plus (`+`) is permitted but ignored.
+- Underscore (`_`) is permitted but ignored, and it is suggested to avoid it.
```
keszybz, line 25: I would suggest the following instead:
```
The version string is a sequence of zero or more characters.

The following characters have special meaning:
- ASCII digits (`0-9`) form numerical components.
- ASCII letters (`a-z`, `A-Z`) form alphabetical components.
- Dot (`.`) separates parts of a component.
- Minus (`-`) separates major parts of the version string.
- Tilde (`~`) starts a suffix that always sorts lower.
- Caret (`^`) starts a suffix that always sorts higher.

Other characters are treated as separators.
This includes plus (`+`) and underscore (`_`) and other printable or non-printable characters.
The underscore MAY be used.
The plus SHOULD NOT be used, to avoid confusion with SEMVER which attaches a special meaning to it.
Other characters MUST NOT be used in a version string.

Note that in some contexts (for example <link to DDI spec>, DEB package file names), the underscore is used as a separator and cannot be used freely in the version string.
```

This changes the text to explicitly say that other characters are disallowed.
Fedora allows `[a-zA-Z0-9.~^_-]` (*), Debian the same except for `_` and `^` (
https://www.debian.org/doc/debian-policy/ch-controlfields.html#source, https://www.debian.org/doc/debian-policy/ch-controlfields.html#version), semver also a subset. So I think we can do this narrowing, because all package name-versions in distributions satisfy this requirement.

(*) The docs in Fedora could certainly be improved.
https://docs.fedoraproject.org/en-US/packaging-guidelines/Naming/#_common_character_set_for_package_naming describes the Name, but https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/ isn't so nice as to list what characters are allowed and is a huge wall of text.

specs/version_format_specification.md
```diff
@@ -103,6 +105,14 @@ Examples (with '' meaning the empty string):
 * `0.0` > `0`
 * `0` < `~`
 * '' < `~`
+* `1_` == `1`
+* `_1` == `1`
+* `1_` < `1.2`
+* `1_2_3` > `1.3.3`
```
keszybz, line 111: I would add an explanation here:
"The underscore is a separator between components, so we first compare "1" with "1.3.3" as numerical version strings, and 1 < 1.3.3. The remainder of the first string is not used in the comparison."

specs/version_format_specification.md
bluca: looks great, updated and added Co-authored-by

specs/version_format_specification.md
```diff
@@ -79,6 +107,18 @@ Examples (with '' meaning the empty string):
...
+* `1+` == `1`
+* `+1` == `1`
+* `1+` < `1.2`
+* `1+2+3` > `1.3.3`
+
+Note how in the `1_2_3` < `1.3.3` case, the underscore and plus characters act as separators between
```
keszybz, line 119: Hmm, "plus" is not used here.

bluca: added

PR #55, opened 2023-03-24
sysext: mention new identification fields

bluca:
As defined in the os-release manpage by:
https://github.com/systemd/systemd/pull/26941


specs/sysext.md
```diff
@@ -43,6 +44,18 @@ Examples: `"SYSEXT_LEVEL=2"`, `"SYSEXT_LEVEL=15.14"`.
 
 If not present, and if `VERSION_ID=` is present instead, then this will be checked instead.
 
+#### `VERSION_ID=` `ID=`
+These are used for exact matches, and the meaning is defined in the
+[`os-release` specification](https://www.freedesktop.org/software/systemd/man/os-release.html).
```
pothos, line 49:
```suggestion
[`os-release` specification](https://www.freedesktop.org/software/systemd/man/os-release.html).
For the `ID=` field the special value `_any` can be set which matches any `os-release` value. The fields `VERSION_ID` or `SYSEXT_LEVEL` are not matched then.
```
Since the line above only talks about exact matches, I think it would be good to avoid conflicting statements/confusion with lines 31-34 by adding this clarification. Otherwise it sounds like `VERSION_ID` and `ID=` would always required to be exact matches.

specs/sysext.md
```diff
@@ -33,7 +33,8 @@ be ignored, while if it is not present, but `VERSION_ID=` is, then the latter mu
 In addition, the `ID=` field must be present and match the base image's, or be set to the special value
 `_any`, in case the sysext can be used on any Linux distribution.
 
-### Fields in extension-
+### Fields in extension-release - Matching with the root/usr DDI
```
keszybz, line 36: `-` →`—` please, here and below.

specs/sysext.md
```diff
@@ -43,6 +44,21 @@ Examples: `"SYSEXT_LEVEL=2"`, `"SYSEXT_LEVEL=15.14"`.
 
 If not present, and if `VERSION_ID=` is present instead, then this will be checked instead.
 
+#### `VERSION_ID=` `ID=` `ARCHITECTURE=`
+`VERSION_ID=` and `ID=` are used to match the sysext with the root/usr DDI, and `ARCHITECTURE=` is used
+to match with the host's CPU architecture, as defined in the
+[`os-release` specification](https://www.freedesktop.org/software/systemd/man/os-release.html).
+`ID=` and `ARCHITECTURE=` also support specifying the `_any` wildcard, which allows to bypass the
+matching mechanism.
```
keszybz, line 52: "which allows the matching mechanism to be bypassed"

specs/sysext.md
```diff
@@ -43,6 +44,21 @@ Examples: `"SYSEXT_LEVEL=2"`, `"SYSEXT_LEVEL=15.14"`.
...
+
+### Fields in extension-release - Identifying the Sysext
+The identification fields defined in the
+[`os-release` specification](https://www.freedesktop.org/software/systemd/man/os-release.html)
+can be used to also identify the sysext itself, by prefixing them with `SYSEXT_`. For example,
+`SYSEXT_ID=myext` `SYSEXT_VERSION_ID=0.1` denotes a 'myext' sysext of version '0.1.
```
keszybz, line 58: Markup here is broken around "0.1".

keszybz (changes requested):
Sorry for the late review.
I wrote this earlier, but couldn't submit, because github was distracted.

bluca:
> Sorry for the late review. I wrote this earlier, but couldn't submit, because github was distracted.

https://github.com/uapi-group/specifications/pull/56

PR #57, opened 2023-04-13
sysext: cover confexts too

bluca:
Rename specification to cover both types of extensions


specs/extension_image.md
```diff
@@ -1,65 +1,79 @@
 ---
-title: Sysext - System Extension
+title: Extension Images
```
keszybz, line 2: The new name is better… But can you provide a redirect from the name to the new? I'm sure that there are links out there to the old ULR.

keszybz (changes requested):

specs/extension_image.md
bluca: done

keszybz (approved):

PR #58, opened 2023-04-13
specs: import the table of PCRs 0—7 from systemd-cryptenroll(8)

keszybz:
This table is informative anyway. If somebody wants to make any realistic use of it, they will usually want a full list of PCRs, because policies bind to e.g. PCR 7 too. So let's copy in the remaining part from systemd-cryptenroll(8). In the preamble it is mentioned that the list here is not authoritative.

In the man page, the discussion of when various PCRs change was included directly in the table. It's tough to squeeze the extra text into the table, so I moved after the table.

I dropped "(cont.d)" for the first row. When the table is rendered, an empty field is visually clearer than the text.

https://sembr.org/ is used for wrapping.


specs/linux_tpm_pcr_registry.md
```diff
@@ -28,6 +30,70 @@ In both cases it is important that data measured into the PCRs is carefully chos
    <th><strong>Log</strong></th>
    <th><strong>Use Reported By</strong></th>
   </tr>
+
+  <tr>
+   <td style="background-color:#fff3bf;"><p style="text-align: right"><strong>0</strong></p></td>
+   <td style="background-color:#AEA;"><code style="background-color:#AEA;">Firmware 💻</code></td>
```
poettering, line 36: the background color seems wrong. should be grey, not green, like for the other firmware entries

keszybz: Search-and-replace hard. I'll submit a follow-up.

PR #60, opened 2023-04-22
UKI: define a location for centrally built UKIs

behrmann:
We were discussing in mkosi how to handle distribution-supplied UKIs in the future. It would be helpful to have a common location that producers can put them in and consumers can look for them. 

I'm open to a different location or naming scheme, the point is rather that there is a standard. 

I didn't touch on the UKI being installed to the ESP or XBOOTLDR directly (in Debian land dpkg can't even do that) - it would be possible within the *should* though -, but I think the general way would be for the package manager to put the UKI in the standard location and kernel-install or a tool like it, to marry it to local configuration (e.g. number of tries for boot counting) and putting it where it should go (ESP or XBOOTLDR).

### DIFF
--- a/specs/unified_kernel_image.md
+++ b/specs/unified_kernel_image.md
@@ -111,3 +111,12 @@ individual resources listed above at once, and their combination. Standard Linux
 [`sbsigntool`](https://manpages.debian.org/unstable/sbsigntool/sbsign.1.en.html) and
 [`pesign`](https://github.com/rhboot/pesign) can be used to sign UKI files. The signature format and process
 again match the ones already used for PE files, so they will not be redefined here.
+
+## Distribution-built UKIs Installed by Package Managers
+
+UKIs that are built centrally by distributions and installed via the package manager should be installed in
+`/usr/lib/modules/$UNAME/`, where `$UNAME` is the output of `uname -r` of the kernel included in the UKI, so
+that tools staging or consuming UKIs have a common place to store and look for them.
+
+The installed UKIs should have a filename `<version format specification>.efi`, i.e. the filename is left to
+implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).


specs/unified_kernel_image.md
```diff
@@ -111,3 +111,9 @@ individual resources listed above at once, and their combination. Standard Linux
 [`sbsigntool`](https://manpages.debian.org/unstable/sbsigntool/sbsign.1.en.html) and
 [`pesign`](https://github.com/rhboot/pesign) can be used to sign UKI files. The signature format and process
 again match the ones already used for PE files, so they will not be redefined here.
+
+## Distribution-built UKIs Installed by Package Managers
+
+UKIs that are built centrally by distributions and installed via the package manager should be installed
```
bluca, line 117: "installed in"

specs/unified_kernel_image.md
```diff
@@ -111,3 +111,9 @@ individual resources listed above at once, and their combination. Standard Linux
...
+`/usr/lib/kernel/ukis/` with a filename `<version format specification>.efi`, so that tools staging or
```
bluca, line 118: I wonder if we should allow multiarch too? Given we don't use the EFI arch suffix for now, they would clash otherwise

specs/unified_kernel_image.md
behrmann: Fixed.

specs/unified_kernel_image.md
behrmann: I don't see a reason why not. Happy to change this, if you have a formulation. The filename scheme was just the first thing I thought of on the spot, since I assumed something like the name `6.1.0-7-amd64` I see on a Debian system in `/usr/lib/modules` could be used to accommodate this within the version format specification.

bluca: yeah we can use the kernel package name which usually has the arch

bluca (approved):

poettering:
Hmm, so since a long time on Fedora kernels shipped in RPM are placed in `/usr/lib/modules/$(uname -r)/vmlinuz`, and that makes a ton of sense, i.e. at the same place as kernel modules, in the per-kernek-version dir. I think we really should stick to that.

I have the suspicion that distros might want to build different flavours of UKIs from the same ELF kernel image, hence I figure we should have a dir below that, and we probably want a dir where initrd sysext images can be placed.

Thus maybe:

`/usr/lib/modules/$(uname -r)/uki/<somekernelflavourname>.efi`

and then:

`/usr/lib/modules/$(uname -r)/uki/<somekernelflavourname>-<somesysextname>.sysext.raw`

or so?

I am pretty sure @keszybz should say something about this, given he's implementing a good chunk of this.

poettering:
(and we probably should standardize the place to put the ELF kernel image too, so that kernel-install can find it automatically without being explicitly specified via an absolute path. i.e. standardize what fedora is already doing)

berrange:
> Hmm, so since a long time on Fedora kernels shipped in RPM are placed in /usr/lib/modules/$(uname -r)/vmlinuz

snip

> Thus maybe:
>
> /usr/lib/modules/$(uname -r)/uki/<somekernelflavourname>.efi

The first introduced UKI for general purpose VMs is shipping in Fedora 38 and CentOS9 Stream today at

```
/usr/lib/modules/$(uname -r)/vmlinuz-<flavourname>.efi
```

ie

```
/lib/modules/6.2.11-300.fc38.x86_64/vmlinuz-virt.efi
```

behrmann:
I think I'd be fine with `/usr/lib/modules/$(uname -r)/vmlinuz-<flavourname>.efi`, although I'd prefer @poettering's proposal, since the `vmlinuz-` name strikes me as a bit weird, because it's not just the kernel image, but the initrd, too, no?

@bluca What do you think about these for Debian `uname -r` does include the the arch on Debian, so that should be fine?

/cc @Foxboron for some Arch input, in case Arch wants to ship UKIs at some point.

bluca:
The location is fine. For the filename, I don't think we need to lock down on one format here? This is just about the stgorage directory, right?

berrange:
> I think I'd be fine with /usr/lib/modules/$(uname -r)/vmlinuz-<flavourname>.efi, although I'd prefer @poettering's proposal, since the vmlinuz- name strikes me as a bit weird, because it's not just the kernel image, but the initrd, too, no?

The fact that the kernel has an initrd grafted onto the side of it is a minor impl detail. From the user POV, these binaries are the kernel images that they are choosing between for booting their system. The bootloader is going to present the user with a list of kernels, which can be either UKIs or traditional split kernel/initrd/cmdline - either way they look broadly the same once presented to the user, so it felt appropriate to have vmlinuz in the name, with the .efi suffix indicating it had the other pieces (initrd/cmdline/etc) grafted on.

Foxboron:
> with the .efi suffix indicating it had the other pieces (initrd/cmdline/etc) grafted on.

I don't quite understand why the `.efi` suffix would indicate any such thing though.

keszybz:
From the original proposal:
> with a filename `<version format specification>.efi`

I like this way of phrasing this, it nicely encompasses various schemes that people are already using. And I don't think it's feasible to define the details of the file name. Different distros will want to have different granularity of versioning and flavouring of the images, so some variance in naming is unavoidable, and this way we avoid the issue. But it is important that those names are compatible with our version format specs, so that they can be unambiguously enumerated and sorted.

>  I'm unsure about `vmlinuz-<flavourname>.efi` as we'd overload the word `vmlinux`.

Right. Let's not specify the filename prefix and sidestep the issue.

> > with the .efi suffix indicating it had the other pieces (initrd/cmdline/etc) grafted on.
> 
> I don't quite understand why the `.efi` suffix would indicate any such thing though.

This is a natural consequence of how we name images. Usually the original binary file has no suffix, and we would append `.efi` to get a different name. The boot loader specification mandates `/EFI/Linux/*.efi`. That's what is described in `dracut(8)` for `--uefi`, and the examples for `ukify(1)` use the same pattern, since it's natural to name the output file under the same name under which it will be installed.    

> My first take is that `/usr/lib/modules/$(uname -r)/` should be fine for the location.

Yeah.

I don't like the `/uki/` part suggested by @poettering. The fact that this is an UKI is something of an implementation detail. And since we already have the `.efi` suffix, we don't need `/uki/` in the path.

There's also the "first mover advantage". All other considerations being equal, it'd be nicer to follow what Fedora does (`/usr/lib/modules/$(uname -r)/vmlinuz-<flavourname>.efi`), since that's already deployed.

Foxboron:
> This is a natural consequence of how we name images. Usually the original binary file has no suffix, and we would append `.efi` to get a different name. The boot loader specification mandates `/EFI/Linux/*.efi`. That's what is described in `dracut(8)` for `--uefi`, and the examples for `ukify(1)` use the same pattern, since it's natural to name the output file under the same name under which it will be installed.

`.efi` is used because that is what you naturally name EFI binaries. I'm question the assertion that this would imply we have the UKI sections available.

behrmann:
Pushed a change incorporating the feedback. 'Changed the location to `/usr/lib/modules/$(uname -r)` and kept the filename as `<version format spec>.efi`. This should accommodate Fedora moving first on that, but leaves it for others to go with a different name, as I'm with @Foxboron to not step on the kernel's toes with the `vmlinuz-` name.

keszybz (approved):

poettering:
> I don't like the `/uki/` part suggested by @poettering. The fact that this is an UKI is something of an implementation detail. And since we already have the `.efi` suffix, we don't need `/uki/` in the path.

So the reason why I wanted to separate this out in a subdir was mostly because of system extensions of the initrd that are pre-built and should be placed adjacent so the kernel somehow, so that we can install them along with the UKI in the ESP when needed. For such an install operation it's nice to have the stuff in a separate dir, so that the installer can be dumb and just copies the whole dir contents over without having to filter out files by some glob or so. 

That said instead of doing `/usr/lib/modules/$(uname -r)/uki/<somekernelflavourname>.efi` and `/usr/lib/modules/$(uname -r)/uki/<somekernelflavourname>-<somesysextname>.sysext.raw` as I proposed earlier we could also do:

`/usr/lib/modules/$(uname -r)/<somekernelflavourname>.efi` and `/usr/lib/modules/$(uname -r)/sysext/<somekernelflavourname>-<somesysextname>.sysext.raw`.

Or maybe:

`/usr/lib/modules/$(uname -r)/<somekernelflavourname>.efi` and `/usr/lib/modules/$(uname -r)/<somekernelflavourname>.efi.sysext.d/<somekernelflavourname>-<somesysextname>.sysext.raw`.

Or so?

In other words: i think a subdir for the kernels themselves is not as important to me, but sysexts should be isolated into some subdir i guess, so that we can easily copy them in one go.

poettering:
btw, I wouldn't mind if we'd also mention a recommended path for the non-UKI kernel images, i.e. the type1 images. or on other words: if we'd standardize the paths fedora already uses for those in the spec, and thus can teach systemd's `kernel-install` script to imply the path to the kernel image so that ultimately a simple `kernel install add <kver>` suffices to regenerate everything.

berrange:
> Or maybe:
>
> /usr/lib/modules/$(uname -r)/<somekernelflavourname>.efi and /usr/lib/modules/$(uname -r)/<somekernelflavourname>.efi.sysext.d/<somekernelflavourname>-<somesysextname>.sysext.raw.

I think that makes decent sense as it means the subdir only need exist if there are sysexts present/required for a given kernel

behrmann:
This has grown a bit now and encompasses multiple topics:
1. central UKI location
2. location for sidecar sysexts
3. possible naming scheme for what's called `vmlinuz` (with some possible suffix) in most distros

I'll move the whole thing into the bootloader spec and add references into the UKI and sysext specs. If anybody has a better idea than putting it in the bootloader spec I'm happy to do that instead.

keszybz:
I don't think this is really a matter for the BLS. The BLS is about how the boot loader finds images, and the location of kernels in the fs is a separate issue. It's also Linux-specific, while BLS is fairly general.

I would like to merge this as-is, and then we can always extend it later with locations for sysexts and such.

poettering:
sounds ok to me.

behrmann:
That's also fine with me.

DaanDeMeyer:
Something of note is that when CONFIG_ZBOOT is used, the resulting kernel image on arm64, loongarch and riscv is named vmlinuz.efi, so we might still be intruding on the kernel's namespace with this approach.

PR #62, opened 2023-05-30
spec: fix tilde comparison examples in version spec

a3f:
[Rule 3](https://uapi-group.org/specifications/specs/version_format_specification/#version-comparison) specifies that:

>   If the remaining part of one of strings starts with ~:
>   if other remaining part does not start with ~, the string with
>   ~ compares lower. Otherwise, both tilde characters are skipped.

The two examples specifying that ~ would compare greater than 0 or the empty string are thus wrong. The discrepancy likely results from systemd changing its comparison algorithm in v252-rc1: https://github.com/systemd/systemd/commit/46083ab321826d920b32de85c7328a5f39b7fff4

Fix the mismatch by aligning the example with the rules and with the systemd implementation of `strverscmp_improved`.


keszybz:
Indeed.

PR #68, opened 2023-08-22
Clarify that /loader/entries should be in the root of the ESP filesystem

DaanDeMeyer:
The current wording sometimes gets misinterpreted as having to search for /loader/entries on all filesystems in the partition table so let's clarify that /loader/entries should be located on the root of the ESP filesystem.

### DIFF
--- a/specs/boot_loader_specification.md
+++ b/specs/boot_loader_specification.md
@@ -170,8 +170,9 @@ only: while boot loaders should look in both places, OSes should only add their
 files to `$BOOT`.
 
 **Note:** _In all cases the `/loader/entries/` directory should be located
-directly in the root of the file system. Specifically, the `/loader/entries/`
-directory should **not** be located under the `/EFI/` subdirectory on the ESP._
+directly in the root of the ESP or XBOOTLDR filesystem. Specifically, the
+`/loader/entries/` directory should **not** be located under the `/EFI/`
+subdirectory on the ESP or XBOOTLDR._
 
 The file name of the boot entry snippets is used for identification of the boot
 item but shall never be presented to the user in the UI. The file name may be


specs/boot_loader_specification.md
```diff
@@ -170,7 +170,7 @@ only: while boot loaders should look in both places, OSes should only add their
 files to `$BOOT`.
 
 **Note:** _In all cases the `/loader/entries/` directory should be located
-directly in the root of the file system. Specifically, the `/loader/entries/`
+directly in the root of the ESP file system. Specifically, the `/loader/entries/`
```
keszybz, line 173: No, because it can be on ESP or on XBOOTLDR. So this phrasing is also wrong.

specs/boot_loader_specification.md
```diff
@@ -170,8 +170,9 @@ only: while boot loaders should look in both places, OSes should only add their
 files to `$BOOT`.
 
 **Note:** _In all cases the `/loader/entries/` directory should be located
-directly in the root of the file system. Specifically, the `/loader/entries/`
-directory should **not** be located under the `/EFI/` subdirectory on the ESP._
+directly in the root of the ESP or `$BOOT` file system. Specifically, the
```
behrmann, line 173: But `$BOOT` can be the ESP, so why not just `$BOOT`?

DaanDeMeyer: Because it can also not be the ESP, and it that case we look on both the ESP and $BOOT

poettering: $BOOT is supposed to indicate where we *put* boot entries. It points to either ESP or XBOOTLDR, depending on the local setup, and it only matters when wondering where to drop new stuff.

If we change the perspective, and are *looking* for boot entries, then $BOOT is not the right word to use, in particular as we actually want to look into both ESP *and* XBOOTLDR always. Hence, let's make this clear here, and list *both* dirs under their real names, not under the singleton pointer to one of them.

Hence, I think this should be rewored to say "root of the ESP or XBOOTLDR partition".

keszybz:
Minor follow-up in https://github.com/uapi-group/specifications/pull/69.

Issue #66, opened 2023-07-19
UKI specification: support multiple embedded device trees


diogo-ivo:
Currently the UKI specification states that there can be one .dtb section where a Device Tree is placed. However, there are use cases where it is useful for one UKI to contain multiple DTs, allowing a single image to support multiple (similar) boards. As such, it would be good to see if there is interest in standardizing this feature.

An initial proposal for this is as follows:

- The extra FDTs are embedded in the UKI by adding sections named .dtb-* for each additional Device Tree (this is already implemented in efibootguard). Another solution would be to concatenate the FDTs in the same .dtb section (with possible appropriate padding), but this raises some problems with maintaining backwards compatibility.
- The selection of the appropriate FDT is done by matching the compatible strings against the firmware provided FDT

bluca:
Yes we want to support this, it needs to be implemented in sd-stub at the same time though

jan-kiszka:
To be clear: Ivo would also contribute the needed code changes to sd-stub. He already has a working prototype, just not yet exactly with the suggested format above.

poettering:
I guess we can close this here, now that #71 has been merged.

(Would have preferred if we merge the systemd PR and this spec PR at the same time, but I guess that ship has sailed)

PR #71, opened 2023-08-24
UKI: document the case of multiple Device Trees

diogo-ivo:
In connection with the discussion in [1] this pull request specifies how the case of multiple Device Trees is to be handled by choosing to append all additional Device Trees in the same .dtb section.

[1]: https://github.com/uapi-group/specifications/issues/66

### DIFF
--- a/specs/unified_kernel_image.md
+++ b/specs/unified_kernel_image.md
@@ -48,10 +48,12 @@ UKIs consist of the following resources:
 * Optionally, information describing the OS this kernel is intended for, in the `.osrel` section. The contents of this section are derived from `/etc/os-release` of the target OS. They can be useful for presentation of the UKI in the boot loader menu, and ordering it against other entries using the included version information.
 * Optionally, information describing kernel release information (i.e. `uname -r` output) in the `.uname` section. This is also useful for presentation of the UKI in the boot loader menu, and ordering it against other entries.
 * Optionally, a splash image to bring to screen before transitioning into the Linux kernel, in the `.splash` section.
-* Optionally, a compiled Devicetree database file, for systems which need it, in the `.dtb` section.
+* Optionally, one or more compiled Device Trees, for systems which need it, each in its separate `.dtb` section. If multiple `.dtb` sections exist then one of them is selected according to an implementation-specific algorithm.
 * Optionally, the public part of a public-private key pair in PEM format used to sign the image, in the `.pcrpkey` section.
 * Optionally, a JSON file encoding expected PCR 11 hash values seen from userspace once the UKI has booted up, along with signatures of these expected PCR 11 hash values, in the `.pcrsig` section. The signatures must also match the abovementioned key pair.
 * Optionally, a CSV file encoding the SBAT metadata for the image, in the `.sbat` section. The [SBAT format is defined by the Shim project](https://github.com/rhboot/shim/blob/main/SBAT.md), and used for UEFI revocation purposes.
+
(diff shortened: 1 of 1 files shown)


specs/unified_kernel_image.md
```diff
@@ -48,7 +48,7 @@ UKIs consist of the following resources:
 * Optionally, information describing the OS this kernel is intended for, in the `.osrel` section. The contents of this section are derived from `/etc/os-release` of the target OS. They can be useful for presentation of the UKI in the boot loader menu, and ordering it against other entries using the included version information.
 * Optionally, information describing kernel release information (i.e. `uname -r` output) in the `.uname` section. This is also useful for presentation of the UKI in the boot loader menu, and ordering it against other entries.
 * Optionally, a splash image to bring to screen before transitioning into the Linux kernel, in the `.splash` section.
-* Optionally, a compiled Devicetree database file, for systems which need it, in the `.dtb` section.
+* Optionally, one or more compiled Device Trees, for systems which need it, each in its separate `.dtb` section.
```
poettering, line 51: Can you add a sentence after this itemized list that says something like this:

"Note that all of the sections defined above are singletons: they may appear once at most – except for the `.dtb` section which may be appear more than once."

poettering: Also, this should say what it means if multiple .dtb sections exist. i.e. that one is picked according to some rule of picking the most appropriate one. I wouldn't define what the algorithm to pick it is, but it certainly deserves mention that only one is picked, and not all.

specs/unified_kernel_image.md
```diff
@@ -48,11 +48,13 @@ UKIs consist of the following resources:
 * Optionally, information describing the OS this kernel is intended for, in the `.osrel` section. The contents of this section are derived from `/etc/os-release` of the target OS. They can be useful for presentation of the UKI in the boot loader menu, and ordering it against other entries using the included version information.
 * Optionally, information describing kernel release information (i.e. `uname -r` output) in the `.uname` section. This is also useful for presentation of the UKI in the boot loader menu, and ordering it against other entries.
 * Optionally, a splash image to bring to screen before transitioning into the Linux kernel, in the `.splash` section.
-* Optionally, a compiled Devicetree database file, for systems which need it, in the `.dtb` section.
+* Optionally, one or more compiled Device Trees, for systems which need it, each in its separate `.dtb` section. If multiple .dtb sections exist then one of them is selected according to an implementation specific algorithm.
```
keszybz, line 51: "implementation-specific" (with dash).

specs/unified_kernel_image.md
```diff
@@ -48,11 +48,13 @@ UKIs consist of the following resources:
...
 * Optionally, the public part of a public-private key pair in PEM format used to sign the image, in the `.pcrpkey` section.
 * Optionally, a signature made using the abovementioned key pair, in the `.pcrsig` section.
 * Optionally, a JSON file encoding expected PCR 11 hash values seen from userspace once the UKI has booted up, along with signatures of these expected PCR 11 hash values, in the `.pcrsig` section. The signatures must also match the abovementioned key pair.
 
+Note that all of the sections defined above are singletons: they may appear once at most – except for the .dtb section which may be appear more than once.
```
keszybz, line 56: Backticks around `.dtb`.

keszybz (changes requested):

specs/unified_kernel_image.md
diogo-ivo: Thank you for the review comments, I have addressed these points in the updated push

bluca (approved):

keszybz (approved):

PR #72, opened 2023-09-09
specs: mark .cmdline in UKI as optional

bluca:
The .cmdline section in a UKI is optional in both sd-stub and ukify, so mark it as such in the specification too


poettering (approved):
migth make sense do say what it is supposed to mean when the `.cmdline` section is absent, and what the difference to including a `.cmdline` but setting it to any empty string is. (i.e. that an absent `.cmdline` means that the cmdline is overridable even in SecureBoot mode).

poettering:
> Isn't that more of a sd-stub detail (that we might want to change separately)? ie it's what the loader does, rather than what the image is

well, sd-stub implements the logic, but I am pretty sure that the spec should dictate what sections (their presence and their absence) are supposed to mean.

bluca:
-> https://github.com/uapi-group/specifications/pull/73

Issue #35, opened 2022-11-01
Sync "Discoverable Partitions Specifications" with "Boot Loader Specification"


ljrk0:
The change https://github.com/systemd/systemd/commit/0b81e47e801c503ba94ccfb04e5372868f9fdf63 also implemented in this repository changes (reverts) the meaning of `$BOOT` as well as the recommended mounting scheme. To cite from the BLS:

> Otherwise [no existing XBOOTLDR partition], if on GPT and an ESP is found and it is large enough (let’s say at least 1G) it should be used as $BOOT and used as primary location to place boot loader menu resources in.

And:

> It is recommended to mount $BOOT to /boot/, and the ESP to /efi/. If $BOOT and the ESP are the same, then either a bind mount or a symlink should be established making the partition available under both paths.

From the commit message:

> 1. I intend to make clear the $BOOT is the "primary" place to put stuff
>   in, and is what should be mounted to /boot/.
>
> 2. The ESP (if different from $BOOT) is listed as "secondary" source to
>    read from, and is what should be mounted to /efi/. NB we now make the
>    distinction between "where to put" (which is single partition) and
>    "where to read from".

This is perhaps a bit less clear than it could:  In the case of ESP being `$BOOT`, should it be mounted to `/boot` or `/efi` and in what "direction" should the symlink/bind-mount be?  While this is probably irrelevant, this spec seems to recommend mounting to `/boot`.  However, the DPS specifies:

> The ESP used for the current boot is automatically mounted to /efi/ (or /boot/ as fallback), unless a different partition is mounted there (possibly via /etc/fstab, or because the Extended Boot Loader Partition — see below — exists) or the directory is non-empty on the root disk. This partition type is defined by the [UEFI Specification](http://www.uefi.org/specifications).
>
> The Extended Boot Loader Partition (XBOOTLDR) used for the current boot is automatically mounted to /boot/, unless a different partition is mounted there (possibly via /etc/fstab) or the directory is non-empty on the root disk. This partition type is defined by the [Boot Loader Specification](https://systemd.io/BOOT_LOADER_SPECIFICATION).

So here the ESP should be mounted to `/efi` by default, and only then to `/boot`.  Further, if a XBOOTLDR partition exists (that is, ESP is *not* `$BOOT`) the spec is unclear whether `/boot` should then be used (which is exactly opposite to the BLS which then would mount the ESP to `/efi` and `$BOOT`/XBOOTLDR to `/boot`) or somewhere else entirely.

[...]

PR #75, opened 2023-09-15
Reconcile the handling of /boot and /efi in BLS and DPS

keszybz:
Fixes https://github.com/uapi-group/specifications/issues/35.

https://github.com/systemd/systemd/commit/6a488fa7cce8124fa885adf8a2f31363fe62f636 updated systemd-gpt-auto-generator as the first step. Now this commit updates both specifications to match the implementation. The implementation is actually much more complex than the specs, because it needs to take into account a state where the mount points are missing, or used by something else, or contain files. The specs just describe the desired end state:

>  It is recommended to mount $BOOT (either XBOOTLDR or the ESP) to /boot/. If both XBOOTLDR and the ESP are present, the ESP should be mounted to /efi/.

> Effectively, this means that /boot/ is the location where new boot entries shall be written to. For the boot loader itself, the situation is more complicated, because its files may be located under /boot/ or /efi/, depending on whether XBOOTLDR exists.

So now the implementation, the man page, and the two specs should be in sync.


DaanDeMeyer:
LGTM

brauner (approved):

poettering:
lgtm

PR #78, opened 2023-09-22
specs/extension_image: Add optional mutability for extensions

t-lo:
This PR extends the extension specification (covering confext and sysext) by optional mutability modes for extensions.

Specifically, it discusses 3 modes:
1. Immutable. This is the default and the way extensions are currently defined.
2. Mutable with a persistent backing directory ("upperdir"). This allows changed to the filesystem's base directories even after extensions are overlaid.
3. Ephemeral. Base directories are mutable, but changes go to ephemeral storage like tmpfs.

A total of 3 configuration options are discussed for setting up mutable modes for base directories.

Please refer to the diff included in this PR for the changes, and /or to the full document at https://github.com/flatcar/uapi-group-specifications/blob/main/specs/extension_image.md .

Prior discussions here: https://github.com/flatcar/Flatcar/issues/986#issuecomment-1571429705 , https://github.com/systemd/systemd/issues/24864#issuecomment-1263311389 .

### DIFF
--- a/specs/extension_image.md
+++ b/specs/extension_image.md
@@ -2,7 +2,7 @@
 title: Extension Images
 category: Concepts
 layout: default
-version: 1
+version: 2
 SPDX-License-Identifier: CC-BY-4.0
 ---
 # Extension Images
@@ -34,9 +34,107 @@ sysext images extend `/usr/` (OS vendor tree) and/or `/opt/` (third-party vendor
 confext images extend `/etc`. They must contain a `/etc/extension-release.d/extension-release.<IMAGE>` file
 to identify them.
 
-## Image Content
-Extension Images should be additive, and not override content present in the base image or other DDIs,
-but this will not be enforced.
+<h2 id="image-content">Image Content</h2>
+Extension Images should be additive, and not override content present in the base image or other DDIs.
+However, there currently is no safe and efficient way to detect collisions and to enforce content uniqueness
+across the stack of images.
+In a future version of this specification options for enforcing uniqueness may be provided.
+
+## Base Directory Immutability / Mutability
+By default, applying ("merging") an Extension Image on a mutable filesystem renders the underlying base
+directory (`/etc/`, `/usr/`, `/opt/`) immutable.
+By implication, merging a confext on a mutable filesystem will result in `/etc/` becoming read-only, and
+merging a sysext might render `/usr/` and/or `/opt/` read-only, depending on the sysext's contents.
+
+This affects base directories actually contained in extensions merged to a root filesystem:
+for instance, if a sysext extends `/usr/` but not `/opt/`, `/opt/` will remain mutable if
+it's on a mutable filesystem.
+However, immutability affects the _full_ base directory.
+Merging an extension on a mutable filesystem that ships a single custom path e.g. below
+`/etc/appconfig-extra/some/sub/path/` will still render the entirety of `/etc/` immutable.
+The base directory or directories remain immutable for as long as extensions are merged.
+Mutability, if present before the merge, is regained only after all extensions overlaying a base directory
+have been un-merged.
+
+### Optional Mutability
+While overlaid base directories are immutable by default, implementations may provide options
+for mutability.
+Retaining mutability may for instance be useful for compatibility with general purpose applications,
+enabling users to operate a "mixed mode" with both system and configuration extensions
+and traditional applications.
+Mixed mode may also be integrated by distributions to facilitate a smooth transition from
+traditional package management to a purely image-based composition of the root file system.
+Lastly, mutability mode may be used on an originally immutable filesystem to allow and to capture
+temporary changes.
+
+### Extension Overlay [Im]Mutability Modes
+System and configuration extensions may operate in one of three modes.
+
+1. *Immutable mode* - The overlaid base directory is immutable.
+    This is the default.
+2. *Mutable mode* - Writes are directed to an _upperdir_ specified by the user or operator
+    (see "Mutability Mode Configuration" below).
+    This _upperdir_ will contain all changes made to the overlaid base directory.
+    1. _Upperdir_ may be specified to _be_ the base directory: may be used to retain mutability
+       of the base directory after extensions have been merged.
+    2. Alternatively, _upperdir_ may be an entirely separate directory: modifications will be captured
+       but the base directory will remain unchanged, retaining its state from before the extension was merged.
+3. *Ephemeral Mode* - Similar to mutable mode (2.) above but writes are only stored temporarily while
+    extensions are merged, and discarded as soon as extensions are un-merged.
+    Location of temporary storage is implementation-specific.
+    Useful for e.g. development and for one-shot validation operations.
+
+### Mutability Mode Configuration
+Immutable mode is the default.
+If none of the configurations outlined below were specified then extension overlays operate
+in immutable mode and base directories are read-only.
+Implementation of any of the below mutable configurations is optional.
+_If_ mutable modes are supported by an implementation, configuration option 1. below _must_ be supported
+for compatibility across implementations.
+
+Mutability modes may be configured in the following ways:
+1. By creating qualified paths or soft-links below `/var/lib/extensions.mutable/`.
+   See "Qualified Paths Definition" below for details.
+   This is the most portable option across different implementations.
+   If an implementation supports base directory mutability then this mode _must_ be supported.
+2. By setting a respective option in an implementation's configuration file.
+   This option is implementation-specific.
+   Implementations may choose to support a single option, multiple options for
+   system and configuration extensions, and/or multiple options per base directory.
+   Using this option should selectively override any qualified path definitions from 1.
+3. By passing a command line parameter upon extension merge or refresh.
+   This option is implementation-specific.
+   Implementations may choose to support a single option and/or multiple options per base directory.
+   Using this option should selectively override any configurations from 1. and 2.
+
+#### Qualified Paths Definition
+
+Mutability Mode 1 enables mutability by creating paths or soft-links below
+`/var/lib/extensions.mutable/`.
+Qualified paths are:
+* `/var/lib/extensions.mutable/etc/` - directory or soft-link to a directory to store writes to
+  `/etc/`. This is for configuration extensions.
+* `/var/lib/extensions.mutable/usr/` - directory or soft-link to a directory to store writes to
+  `/usr/`. This is for system extensions.
+* `/var/lib/extensions.mutable/opt/` - directory or soft-link to a directory to store writes to
+  `/opt/`. This is for system extensions.
+
+Each base directory is treated separately.
+The existence and the type of each qualified path determines the mutability mode used.
+The following mutability modes are supported:
+* _Path does not exist_ - immutable mode.
+* Path is a _directory, subvolume, or mount point_ - the path at
+  `/var/lib/extensions.mutable/<basedir>/` is used to store writes to `/<basedir>/`.
+  * A tmpfs mount at the qualified path may be used for a custom ephemeral mode.
+    In this case, clean-up of the tmpfs is left to the user and/or is implementation-specific.
+* Path is a _soft link_ - the soft link is followed and writes are stored at the link's destination.
+  * If the destination is the base directory - i.e. `/var/lib/extensions.mutable/<basedir>/`
+    points to `/<basedir>/` - then the stacking order changes and `<basedir>` becomes _upperdir_.
+    Writes are directed to the base directory and files and paths present in the base directory
+    override files and paths in extensions if present.
+  * The soft link may point to a tmpfs destination for custom ephemeral mode.
+    In this case, clean-up of the tmpfs is left to the user and/or is implementation-specific.
+  * If the destination does not exist, immutable mode is used.
 
 ## File Suffix
 Since extensions images are DDIs, they should carry the `.raw` suffix. In order to make discerning system


poettering:
(as mentioned elsewhere: let's wait until an implementation of the spec exists. specs made out of thin air that haven't seen an implementation yet, are typically not the highest quality)

t-lo:
> (as mentioned elsewhere: let's wait until an implementation of the spec exists. specs made out of thin air that haven't seen an implementation yet, are typically not the highest quality)

Makes sense. We'll take a stab at adding it to systemd-sysext then.

We'd like to keep this PR open during implementation so we have a place for people to read up on what we want to do.

keszybz (changes requested):

specs/extension_image.md
```diff
@@ -29,8 +29,108 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+However, there currently is no safe and efficient way to detect collisions and to enforce content uniqueness
+across the stack of images.
+In a future version of this specification options for enforcing uniqueness may be provided.
+
+## Base Directory Immutability / Mutability
+The contents of system and configuration extension images are immutable and cannot be changed after the
+image was created.
+By default, the base directory (`/etc`, `/usr`, `/opt`) overlaid by an extension is also rendered immutable
+when the extension is applied ("merged").
+This only affects base directories actually contained in extensions merged to a root FS: for instance, if a
```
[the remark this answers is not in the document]
t-lo: Updated, thanks!

specs/extension_image.md
```diff
@@ -29,8 +29,108 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+sysext extends `/usr` but not `/opt`, `/opt` will remain mutable.
+However, immutability affects the _full_ base directory.
+An extension shipping a single custom path e.g. below `/etc/appconfig-extra/some/sub/path` will still render
+the entirety of `/etc` immutable.
+The base directory or directories remain immutable for as long as the extension(s) are merged.
```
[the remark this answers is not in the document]
t-lo: You're right. I've updated the section.

specs/extension_image.md
```diff
@@ -29,8 +29,108 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+`/usr` and / or `/opt` read-only, depending on the sysext's contents.
+
+### Optional Mutability
+While overlaid base directories are immutable by default, implementations may provide options for retaining
+mutability.
+Retaining mutability may for instance be useful for compatibility with general purpose applications,
+enabling users to operate a "mixed mode" with both system / configuration extensions and traditional
+applications.
+Mixed mode may also be integrated by distributions to facilitate a smooth transition from traditional
+package management to a purely image-bases composition of the root file system if taking such a direction
```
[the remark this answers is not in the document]
t-lo: Nice catch - that typo was too elaborate for vim's spell checker. Fixed.

specs/extension_image.md
```diff
@@ -29,8 +29,108 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+    This _upperdir_ will contain all changes made to the overlaid base directory.
+    It also changes the stacking order compared to immutable mode in that files and paths present
+    in _upperdir_ will override files and paths that are also present in an extension.
+    1. _Upperdir_ may be specified to _be_ the base directory: may be used to retain mutability
+       of the base directory after extensions were merged.
+    2. Alternatively, _upperdir_ may be an entirely separate directory: modifications will be captured
+       but the base directory will remain unchanged, retaining its state from before the extension was merged.
+3. *Ephemeral Mode* - Similar to mutable mode (2.) above but writes are explicitly stored
+    in a temporary directory (possibly in RAM) and discarded as soon as extensions are un-merged.
+    Useful for e.g. development and for one-shot validation operations.
```
krnowak, line 81: From the document I can assume that the `Ephemeral Mode` is completely implementation-defined. There is no way to create an ephemeral layer using the configuration option 1. Is my understanding correct? Might be useful to spell it out explicitly somewhere.

t-lo: Your understanding is correct; I'll add a note on that.

specs/extension_image.md
```diff
@@ -35,8 +35,109 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+the image was created.
+By default, the base directory (`/etc`, `/usr`, `/opt`) overlaid by an extension is also rendered immutable
+when the extension is applied ("merged").
+This only affects base directories actually contained in extensions merged to a root filesystem:
+for instance, if a sysext extends `/usr` but not `/opt`, `/opt` will remain mutable.
+However, immutability affects the _full_ base directory.
+An extension shipping a single custom path e.g. below `/etc/appconfig-extra/some/sub/path` will still
+render the entirety of `/etc` immutable.
+The base directory or directories remain immutable for as long as the extensions are merged.
+Original state is regained only after all extensions overlaying a base directory were un-merged.
```
keszybz, line 54: s/were/have been/ ? I'm not a native speaker, but "were" sounds to me like it must be in the past, and this is potentially in the future.

specs/extension_image.md
```diff
@@ -35,8 +35,109 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
 ## Image Content
-Extension Images should be additive, and not override content present in the base image or other DDIs,
-but this will not be enforced.
+Extension Images should be additive, and not override content present in the base image or other DDIs.
+However, there currently is no safe and efficient way to detect collisions and to enforce content uniqueness
+across the stack of images.
+In a future version of this specification options for enforcing uniqueness may be provided.
+
+## Base Directory Immutability / Mutability
+The contents of system and configuration extension images are immutable and cannot be changed after
```
keszybz, line 44: I think this should say "by default" or otherwise give a hint that it's not always true.

specs/extension_image.md
```diff
@@ -35,8 +35,109 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+
+### Optional Mutability
+While overlaid base directories are immutable by default, implementations may provide options
+for retaining mutability.
+Retaining mutability may for instance be useful for compatibility with general purpose applications,
+enabling users to operate a "mixed mode" with both system / configuration extensions
+and traditional applications.
+Mixed mode may also be integrated by distributions to facilitate a smooth transition from
+traditional package management to a purely image-based composition of the root file system
+if taking such a direction is desired.
```
keszybz, line 67: Strike "if taking such a direction is desired"? It doesn't add any value.

specs/extension_image.md
```diff
@@ -35,8 +35,109 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+    in the base directory.
+2. *Mutable mode* - Writes are directed to an _upperdir_ specified by the user / operator
+    (see "Mutability Mode Configuration" below).
+    This _upperdir_ will contain all changes made to the overlaid base directory.
+    It also changes the stacking order compared to immutable mode in that files and paths present
+    in _upperdir_ will override files and paths that are also present in an extension.
+    1. _Upperdir_ may be specified to _be_ the base directory: may be used to retain mutability
+       of the base directory after extensions were merged.
+    2. Alternatively, _upperdir_ may be an entirely separate directory: modifications will be captured
+       but the base directory will remain unchanged, retaining its state from before the extension was merged.
```
keszybz, line 84: Hmm, so we have two (2.), so the reference below is ambiguous. Please change the inner list to letters or roman numberals or something.

specs/extension_image.md
```diff
@@ -35,8 +35,109 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+3. *Ephemeral Mode* - Similar to mutable mode (2.) above but writes are explicitly stored
+    in a temporary directory (possibly in RAM) and discarded as soon as extensions are un-merged.
```
keszybz, line 86: Tmpfs is swappable, so it's not actually "RAM". I think it's better to say "in a temporary file system" and nothing more.

specs/extension_image.md
```diff
@@ -35,8 +35,109 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+    Location of temporrary storage is implementation specific.
```
keszybz, line 87: Typo in "temporary".

keszybz (commented):
Hmm, I agree with Lennart that we need an implementation for this. This text contains a lot of very specific details.

specs/extension_image.md
keszybz: Yeah: "you can go out after you have done your homework", and not "you can go out after you did your homework".

t-lo: Will fix, happened in multiple places.

specs/extension_image.md
t-lo: No, the _extension contents_ are always immutable. The whole sentence is misleading though since that section discusses base directory mutability. I'll remove it.

specs/extension_image.md
t-lo: It will be rendered into a sub-list because it's indented (which even happens to use roman numerals), but I sure can make this more explicit in the source.

specs/extension_image.md
t-lo: I'll change it to "will only be stored temporarily while sysexts are merged". I'd like to avoid implementation details and I feel my phrasing above is too specific already.

specs/extension_image.md
t-lo: :sweat_smile:

specs/extension_image.md
```diff
@@ -35,8 +35,109 @@ confext images extend `/etc`. They must contain a `/etc/extension-release.d/exte
...
+traditional package management to a purely image-based composition of the root file system
+if taking such a direction is desired.
+
+### Extension Overlay [Im]Mutability Modes
+System and configuration extensions may operate in one of three modes.
+
+1. *Immutable mode* - The overlaid base directory is rendered immutable.
+    This is the default.
+    Files and paths in extensions will by implication override files and paths also present
+    in the base directory.
```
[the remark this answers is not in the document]
t-lo: Yupp, makes sense. Also, it's an implementation detail of (the current version of) overlayfs. Will remove.

t-lo:
Addressed all feedback, thanks @keszybz and @krnowak for the review!

FYI there's a WIP PR for the above (mutability mode #1 only) now: https://github.com/systemd/systemd/pull/31000

specs/extension_image.md
t-lo: I actually needed to revert this because it turns out that when you use 

i.
ii.

it doesn't render as list at all (it's just plain text). If you use indented numbers like above, they render as roman numerals :shrug: 
image

krnowak (commented):
Question: Should the specification say anything about implementation's behavior when there are zero extension images, but there is a mutable layer configured?

pothos:
With @poettering we discussed that `merge` could be changed to always set up the mount for consistent behavior.

specs/extension_image.md
```diff
@@ -34,9 +34,103 @@ sysext images extend `/usr/` (OS vendor tree) and/or `/opt/` (third-party vendor
...
-Extension Images should be additive, and not override content present in the base image or other DDIs,
-but this will not be enforced.
+<h2 id="image-content">Image Content</h2>
+Extension Images should be additive, and not override content present in the base image or other DDIs.
+However, there currently is no safe and efficient way to detect collisions and to enforce content uniqueness
+across the stack of images.
+In a future version of this specification options for enforcing uniqueness may be provided.
+
+## Base Directory Immutability / Mutability
+By default, the base directory (`/etc`, `/usr`, `/opt`) overlaid by an extension is rendered immutable
```
poettering, line 44: in systemd we have this rule that paths in docs and log msgs that reference a dir should always be suffixed with `/` to make this clear. I think it would make sense to follow the same rule here?

specs/extension_image.md
```diff
@@ -34,9 +34,103 @@ sysext images extend `/usr/` (OS vendor tree) and/or `/opt/` (third-party vendor
...
+### Mutability Mode Configuration
+Immutable mode is the default.
+If none of the configurations outlined below were specified then extension overlays operate
+in immutable mode and base directories are read-only.
+Implementation of any of the below mutable configurations is optional.
+_If_ mutable modes are supported by an implementation, configuration option 1. below _must_ be supported
+for compatibility across implementations.
+
+Mutability modes may be configured in the following ways:
+1. By creating qualified paths or soft-links below `/var/lib/extension-data/`.
```
poettering, line 92: what's the rationale for this dir? it sounds weirdly redundant with its "-data" suffix. That doesn't tell you anything what this is.

I originally assumed the mutable layer would be just /usr.local/. Here you are suggesting an entirely different dir however. what's your rationale for this?

So I think there might be value in placing this in /var/lib/. My thinking has evolved a bit on this. Specifically: I think we should keep the option open to one day have "systemd-sysext commit" and "systemd-confext commit", which would take the current mutable layer and automatically turn it into an proper extension (i.e. an immutable layer), and then opening a new mutable layer. the operation for that would be very simple, just move/rename the upper layer dir, so that it becomes a lower layer dir, mkdir a new upper layer dir, and replace the overlayfs mount with the new arrangement.

Hence, from that angle there's value in having both the immutable sysexts/confexts and the mutable layers on the same fs, so that the move/rename can safely work.

Right now systemd suggests putting confexts in /var/lib/confexts/ and sysexts into /bar/lib/extensions/. I think we should sooner or later probably move the latter to /var/lib/sysexts/ actually. But maybe taking inspiration from that the mutable layer should be:

/var/lib/confexts.mutable/
/var/lib/sysexts.mutable/

maybe?

opinions?

specs/extension_image.md
```diff
@@ -34,9 +34,103 @@ sysext images extend `/usr/` (OS vendor tree) and/or `/opt/` (third-party vendor
...
+
+Mutability Mode 1 enables mutability by creating paths or soft-links below
+`/var/lib/extension-data/`.
+Qualified paths are:
+* `/var/lib/extension-data/etc.local` - directory or soft-link to a directory to store writes to
+  `/etc`. This is for configuration extensions.
+* `/var/lib/extension-data/usr.local` - directory or soft-link to a directory to store writes to
+  `/usr`. This is for system extensions.
+* `/var/lib/extension-data/opt.local` - directory or soft-link to a directory to store writes to
+  `/opt`. This is for system extensions.
```
poettering, line 116: Well, with the discussion above I am pretty sure that mutable extensions should look exactly like regular extensions, so that we can turn them into them. Or in other words, the stuff that shows up under /etc/ should be in a dir called "etc", not under one called "etc.local".

Or in other words:

/var/lib/sysexts.mutable/opt/
/var/lib/sysexts.mutable/usr/
/var/lib/confexts.mutable/etc/

are the actually populated dirs?

specs/extension_image.md
poettering: other idea would be to put the mutable layer into /var/lib/confexts/mutable/

But I am not sure how much I like that, because it means there would suddenly be a magic layer whose name doesn't really communicate that it's special.

Maybe /var/lib/confexts/@mutable/? 

@keszybz, @t-lo, @pothos, @bluca opinions?

bluca: Let's not nest and make the whole thing more complicated. A sibling directory in /var/lib/, explicitly called "mutable" or writable or whatever, sounds best to me.

specs/extension_image.md
bluca: I never liked the .local suffix either, so that looks better

specs/extension_image.md
pothos: I like how descriptive the `/var/lib/confexts.mutable/etc/` and ` /var/lib/sysexts.mutable/usr/` paths are.

As for the rename, maybe keep it consistent and stick to `/var/lib/extensions.mutable/usr/` for now?

specs/extension_image.md
```diff
@@ -34,9 +34,103 @@ sysext images extend `/usr/` (OS vendor tree) and/or `/opt/` (third-party vendor
...
+This only affects base directories actually contained in extensions merged to a root filesystem:
+for instance, if a sysext extends `/usr` but not `/opt`, `/opt` will remain mutable.
+However, immutability affects the _full_ base directory.
+An extension shipping a single custom path e.g. below `/etc/appconfig-extra/some/sub/path` will still
+render the entirety of `/etc` immutable.
+The base directory or directories remain immutable for as long as the extensions are merged.
+Original state is regained only after all extensions overlaying a base directory have been un-merged.
+
+By implication, merging a confext will result in `/etc` becoming read-only, and merging a sysext might
+render `/usr` and/or `/opt` read-only, depending on the sysext's contents.
```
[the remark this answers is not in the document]
t-lo: That's a good point, will amend the paragraph.

specs/extension_image.md
t-lo: The proposed directories are quite self-explanatory, I like it. I like Kai's proposal best as it's quite generic. Will amend the PR.

specs/extension_image.md
t-lo: Good point, will update using Kai's proposal.

specs/extension_image.md
t-lo: Resolving. TIL about the oddities of markdown :)

specs/extension_image.md
t-lo: I've updated the PR - makes sense to better distinguish between files and directories.

specs/extension_image.md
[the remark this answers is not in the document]
t-lo: Done.

t-lo:
Updated the PR and addressed all feedback. Mutability is not assumed anymore by the spec.

I've left the discussion around directories backing mutable storage open, but this should be addressed now, too: I went with `/var/lib/extensions.mutable/...` to match the existing `/var/lib/extensions/...`.

@poettering @bluca @pothos would you be available for another review round?

poettering:
lgtm

TriMoon:
From a fast skim of this topic, i understand that you guys want to make the merged content become R/W right?

What about [#31392](https://github.com/systemd/systemd/issues/31392) :thinking:

t-lo:
@TriMoon That's correct - though only optionally (i.e. when explicitly requested). See https://github.com/systemd/systemd/pull/31000 :wink:

krnowak:
Should there be a sentence or a paragraph about implementation being free to gate the mutability mode in a way it wants? Like through additional command-line flags or configuration files? Because right now I'm not sure if the spec says "if `/var/lib/extensions.mutable/usr/` exists, then resulting merged hierarchy must be mutable". Currentlyhttps://github.com/systemd/systemd/pull/31000 gates mutability behind existence of the path AND the command line flag.

poettering:
btw, where are we with this? is this good to merge? does the stuff merged in systemd and this spec match up 100% right now?

t-lo:
Systemd currently implements the symlink configuration option (config option #1, the only mandatory option in the spec). While changes have been merged in systemd, I was planning to conclude this PR only after a new systemd release includes the new feature (this would be systemd-256 if everything works out).

krnowak:
[There was a question that came up during one review](https://github.com/systemd/systemd/pull/31801#discussion_r1560920995) whether confext mutable directory should be `/var/lib/confexts.mutable/`. I'm not sure about this one, but systemd-sysext expects sysexts to be in `/var/lib/extensions/` and confexts in `/var/lib/confexts/`, so possibly for consistency we should have `/var/lib/extensions.mutable/{usr,opt}/` and `/var/lib/confexts.mutable/{etc}/`.

poettering:
Where are we with this now? I think everything in systemd got merged that was suppsoed to be merged?

t-lo:
Thanks for the ping, I'll do one more pass later this week to catch up with any new feedback. Otherwise we should be good.

Would greatly appreciate input from @bluca and @keszybz on whether this is good to merge.

specs/extension_image.md
```diff
@@ -34,9 +34,107 @@ sysext images extend `/usr/` (OS vendor tree) and/or `/opt/` (third-party vendor
...
+  `/opt/`. This is for system extensions.
+
+Each base directory is treated separately.
+The existence and the type of each qualified path determines the mutability mode used.
+The following mutability modes are supported:
+* _Path does not exist_ - immutable mode.
+* Path is a _directory, subvolume, or mount point_ - the path at
+  `/var/lib/extensions.mutable/<basedir>/` is used to store writes to `/<basedir>/`.
+  * A tmpfs mount at the qualified path may be used for a custom ephemeral mode.
+    In this case, clean-up of the tmpfs is left to the user and/or is implementation specific.
```
keszybz, line 129: "implementation-specific" (with a hyphen), here and below.

keszybz (approved):
LGTM.

keszybz:
I think we can merge it. systemd-256 hasn't been released yet, but we're in late rcs, and it's not going to change substantially.

keszybz:
Something went wrong with that push. I see a merge commit.

t-lo:
Yes, that's github's "sync fork to upstream" button breaking stuff. I hate it's not doing a rebase... Will fix.

t-lo:
Aaah, that's better. Also changed `implementation specific` => `implementation-specific`.

PR #79, opened 2023-09-23
extensions: define ordering as implementation dependent

bluca:
sd-sysext uses versioned ordering, while ExtensionImages= respects the specified order


keszybz (commented):
Hmm, do we really want to do this? Shouldn't we just adjust the implementations to use the same rules?

bluca:
Not sure, to me it makes sense that it you specify a specific order because the use case allows it, it is respected, but in other cases you have autodiscovery

keszybz:
Then I'd prefer to say that. "Implementation-defined" ~= "undefined". I think it's much better to say "The default order is …, but implementations may allow a different order to be specified."

keszybz (approved):
Thanks. I think this is crystal clear now.

PR #80, opened 2023-09-23
specs: implement standalone config files specification

bluca:
Merging these patterns into XDG did not work out, so let's define the rules independently for stateless and hermetic-usr configs.

The purpose of this document is to give answers to developers on what the basic patterns need to be in order to support hermetic-usr and stateless systems. First and foremost the recommendation is to use libeconf wherever possible, as that already implements what's necessary and we know it works. If that is not possible, then basic rules over precedence between /etc/ and /usr/ are defined, along with drop-ins and masking. Defining all the low-level nuances as implemented by projects like systemd for their specific use cases is not a goal, as these details are intentionally left for the implementations. Defining what end users need to do or expect is also not a goal, as each application's documentation should be consulted instead.


specs/base_directory_specification.md
```diff
@@ -1,224 +1,63 @@
...
+In order to allow shipping system defaults owned by the OS vendor, while at the same
+time letting local users or admins override those defaults, `/usr/` and `/etc/` are both
+supported for storage of configuration files, with the latter having higher priority.
+Programs must work correctly if no configuration files are found in `/etc/`.
+Optionally, `/run/` is also supported for ephemeral overrides.
+
+For example, `/usr/foo/bar.conf` provides the default configuration file, but a user
+can create `/etc/foo/bar.conf` which would take precedence and completely override
+the former. If `/run/foo/bar.conf` is present and supported, it would take precedence
+over both.
```
jluebbe, line 40: This differs from the documented order for [systemd.unit files](https://www.freedesktop.org/software/systemd/man/systemd.unit.html) at least for drop-ins:

> In addition to /etc/systemd/system, the drop-in ".d/" directories for system services can be placed in /usr/lib/systemd/system or /run/systemd/system directories. Drop-in files in /etc/ take precedence over those in /run/ which in turn take precedence over those in /usr/lib/.

The order for unit files is more complex, but `/etc/systemd/system/*` is searched before `/run/systemd/system/*`, which is searched before `/usr/lib/systemd/system/*`. 

libeconf doesn't seem to support `/run`:

- https://opensuse.github.io/libeconf/libeconf_8h.html#a7ff92b0fb657aa21220a89a0fcb59227
- https://github.com/openSUSE/libeconf/issues/81

Perhaps it would be useful do explain why the proposed order is different/better than what is currently implemented in the referenced projects.

bluca: clerical error

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,64 @@
...
+directories.
+
+## Storage Directories and Overrides
+In order to allow shipping system defaults owned by the OS vendor, while at the same
+time letting local users or admins override those defaults, `/usr/` and `/etc/` are both
+supported for storage of configuration files, with the latter having higher priority.
+Programs must work correctly if no configuration files are found in `/etc/`.
+Optionally, `/run/` is also supported for ephemeral overrides.
+
+For example, `/usr/foo/bar.conf` provides the default configuration file, but a user
```
poettering, line 37: nope. /usr/lib/foo/bar.conf

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,64 @@
...
+can create `/etc/foo/bar.conf` which would take precedence and completely override
+the former. If `/run/foo/bar.conf` is present and supported, it would take precedence
+over `/usr/foo/bar.conf`, but not `/etc/foo/bar.conf`, which would still have the
+highest priority.
+
+## Masking
+As a special override case, it must be possible to mask files across different
+locations by creating a symlink to `/dev/null` or an empty file.
+
+For example, an empty `/etc/foo/bar.conf` means that `/usr/foo/bar.conf` is
```
poettering, line 47: ditto

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Drop-ins
+All configuration paths must support drop-ins. Supporting drop-ins means that in
+addition to parsing a full configuration file, an implementation also must parse all
+files in the drop-in directory associated with it, and merge the result in lexicographic
+order. Drop-ins are not full replacements, but must allow specifying a subset of settings
+present in the main configuration files.
+
+For example, in addition to parsing `/etc/foo/bar.conf`, `/etc/foo/bar.conf.d/a.conf` and
+`/etc/foo/bar.conf.d/b.conf` must also be parsed, in this order.
```
poettering, line 58: this should make clear somehow that /etc/foo/bar.conf.d/*.conf is loaded, i.e. not just a and b, but *everything*

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,64 @@
...
+especially beneficial for image-based deployments, where the vendor tree is read-only.
+
+These rules are derived from existing real-world usage from the [systemd
+project](https://github.com/systemd/systemd) and the [libeconf
+project](https://github.com/openSUSE/libeconf).
+
+This specification is agnostic toward the actual format and content of the configuration
+files, the precise path used under each top-level hierarchy, and also toward the
+filenames and extensions, with the exception of the `.d/` notation for drop-ins
+directories.
```
poettering, line 28: there should be a note somewhere that people really should enforce suffixes on the main config file, and the drop-ins.

poettering: (simply so that we never are confused by ~bak and dpkg-old files and such

specs/configuration_files_specification.md
bluca: previous paragraph says `an implementation also must parse all files in the drop-in directory` which should be clear, in the example here I also wanted to exemplify the ordering

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,64 @@
...
+The purpose of the rules defined here is to allow OS vendors to implement the
+hermetic-usr pattern, where all vendor files are shipped in the vendor tree itself
+(`/usr/`), including configuration files with system defaults, while allowing local
+or vendor overrides without modifying the original files, for easier management. This is
+especially beneficial for image-based deployments, where the vendor tree is read-only.
+
+These rules are derived from existing real-world usage from the [systemd
+project](https://github.com/systemd/systemd) and the [libeconf
+project](https://github.com/openSUSE/libeconf).
+
```
poettering, line 24: hmm, I think this should focus on two distinct cases:

1. the case where a single object (for example a daemon) is configured via a main config file for that single object + drop-ins. example: logind.conf for systemd-logind.

2. the case where multiple objects are configured via only-drops without any "main config file". example: tmpfiles.d/ sysusers.d/ and such.

The semantics of bot are of course very similar, but the absence of the "main config" file in the latter case, is important to mention.

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,64 @@
...
+The purpose of the rules defined here is to allow OS vendors to implement the
+hermetic-usr pattern, where all vendor files are shipped in the vendor tree itself
+(`/usr/`), including configuration files with system defaults, while allowing local
+or vendor overrides without modifying the original files, for easier management. This is
+especially beneficial for image-based deployments, where the vendor tree is read-only.
+
+These rules are derived from existing real-world usage from the [systemd
+project](https://github.com/systemd/systemd) and the [libeconf
+project](https://github.com/openSUSE/libeconf).
```
poettering, line 23: i don't think libeconf implements what systemd implements btw. there are similarities, but it seems to be sufficiently different. they seem to map `/etc/foobar.conf` to `/usr/<vendor>/foobar.conf`, while we map `/etc/<package>/foobar.conf` to `/usr/lib/<package>/foobar.conf`.

Frankly, I don't think what libeconf does makes much sense? Or maybe their docs are just bad? dunno.

poettering: Also, they don't do /run/?

bluca: There is no 100% overlap, but that's fine, because interoperability is not needed. The point here is to have something to point developers to that supports the common patterns of /etc -> /usr, drop-ins and masking.

bluca: The reason for mentioning libeconf is that it's a library that can be integrated anywhere, that people can just use. systemd doesn't provide anything like that.

bluca: > Also, they don't do /run/?

yes, it's mentioned as optional here

jluebbe: > Also, they don't do /run/?

Yes, I noticed that as well (see https://github.com/uapi-group/specifications/pull/80#discussion_r1336284179).

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,74 @@
...
+For example, settings in `/usr/lib/foo/bar.conf.d/a.conf` must override settings in
+`/etc/foo/bar.conf`.
+
+### Drop-ins without Main Configuration File
+Optionally, schemes where only drop-ins, without a 'main' configuration file, should also
+be supported by implementations. In such schemes many drop-ins are loaded from a common
+directory in each hierarchy.
+
+For example, `/usr/lib/foo/a.conf`, `/usr/lib/foo/b.conf` and `/etc/foo/c.conf` are all
+loaded and parsed in this scheme.
```
jluebbe, line 74: In systemd, this would be named:
```suggestion
For example, `/usr/lib/foo.d/a.conf`, `/usr/lib/foo.d/b.conf` and `/etc/foo.d/c.conf` are all
loaded and parsed in this scheme.
```

apt uses `preferences.d`, `sources.list.d` and `trusted.gpg.d`.

Should the directory have a `.d` suffix for consistency?

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,74 @@
...
+time letting local users or admins override those defaults, `/usr/` and `/etc/` are both
+supported for storage of configuration files, with the latter having higher priority.
+Programs must work correctly if no configuration files are found in `/etc/`.
+Optionally, `/run/` is also supported for ephemeral overrides.
+
+For example, `/usr/lib/foo/bar.conf` provides the default configuration file, but a user
+can create `/etc/foo/bar.conf` which would take precedence and completely override
+the former. If `/run/foo/bar.conf` is present and supported, it would take precedence
+over `/usr/lib/foo/bar.conf`, but not `/etc/foo/bar.conf`, which would still have the
+highest priority.
```
jluebbe, line 43: Perhaps follow the order of precedence?

```suggestion
For example, `/usr/lib/foo/bar.conf` provides the default configuration file. 
If `/run/foo/bar.conf` is present and supported, it would take precedence
over `/usr/lib/foo/bar.conf`.
Finally, a user can create `/etc/foo/bar.conf` which would take precedence
and completely override both.
```

specs/configuration_files_specification.md
thkukuk: > i don't think libeconf implements what systemd implements btw. there are similarities, but it seems to be sufficiently different. they seem to map `/etc/foobar.conf` to `/usr/<vendor>/foobar.conf`, while we map `/etc/<package>/foobar.conf` to `/usr/lib/<package>/foobar.conf`.

There is not only `/usr/lib`, but packages also use `/usr/share`  and other directories. Since our goal was to be able to use libeconf for as many existing packages as possible, we decided to not use a fix prefix, but let the application tell libeconf where it has the config files below `/usr`. `<vendor>`  for `lib/<package>`, `share/<package>` or whatever else they are using may not be the best "naming", but the usage of that name has historic reasons.
If somebody has a better naming idea, I have no idea to let the documentation change.

thkukuk: > > Also, they don't do /run/?
> 
> Yes, I noticed that as well (see [#80 (comment)](https://github.com/uapi-group/specifications/pull/80#discussion_r1336284179)).

Support for `/run` was not necessary in the past for our use cases. But it's on my wishlist, there are only no free resources currently.
If we have a spec for configuration files, I see no problem to adjust libeconf to follow that spec.

thkukuk: > > Also, they don't do /run/?
> 
> Yes, I noticed that as well (see [#80 (comment)](https://github.com/uapi-group/specifications/pull/80#discussion_r1336284179)).

Wasn't necessary in the past for our use cases, but it's on my wishlist to get closer to what systemd is doing.

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,74 @@
...
+directories. It is strongly encouraged to enforce a specific suffix for the configuration
+files, in order to disambiguate (e.g.: with backup files), but the choice of which suffix
+to use is left to each implementation.
+
+## Storage Directories and Overrides
+In order to allow shipping system defaults owned by the OS vendor, while at the same
+time letting local users or admins override those defaults, `/usr/` and `/etc/` are both
+supported for storage of configuration files, with the latter having higher priority.
+Programs must work correctly if no configuration files are found in `/etc/`.
+Optionally, `/run/` is also supported for ephemeral overrides.
```
behrmann, line 37: Another thing that came up last year at the summit was, that systemd also supports `/usr/local/lib` for units with precedence in between `/usr/lib/` and `/run/`. It was decided to leave this out because some people did `/usr/lib/etc` (or was it `/usr/share/etc`?). Also mentioned, while not strictly config, at least systemd generators have search paths where `/run/` comes before `/etc/` and the runtime config they generate can have higher precedence than `/etc/` and lower precedece than `/usr/lib/` and systemctl drops transient runtime overrides into `/run/systemd/system.control` that have higher priority than the config in `/etc/systemd/system`.

These are all special cases of one specific project, but other people might have similar needs, so should we maybe give some guidance here, so that if people (probably wrongly) think they have the need to wedge something in between `/usr/lib/` and `/etc/` or (maybe more realistically) need runtime config that trumps config in `/etc/` they arrive at the some solutions as others?

specs/configuration_files_specification.md
bluca: > > i don't think libeconf implements what systemd implements btw. there are similarities, but it seems to be sufficiently different. they seem to map `/etc/foobar.conf` to `/usr/<vendor>/foobar.conf`, while we map `/etc/<package>/foobar.conf` to `/usr/lib/<package>/foobar.conf`.
> 
> There is not only `/usr/lib`, but packages also use `/usr/share` and other directories. Since our goal was to be able to use libeconf for as many existing packages as possible, we decided to not use a fix prefix, but let the application tell libeconf where it has the config files below `/usr`. `<vendor>` for `lib/<package>`, `share/<package>` or whatever else they are using may not be the best "naming", but the usage of that name has historic reasons. If somebody has a better naming idea, I have no idea to let the documentation change.

Ok I'll leave the exact location under `/usr/` up to the implementation, it is a minor detail after all

bluca: > > > Also, they don't do /run/?
> > 
> > 
> > Yes, I noticed that as well (see [#80 (comment)](https://github.com/uapi-group/specifications/pull/80#discussion_r1336284179)).
> 
> Support for `/run` was not necessary in the past for our use cases. But it's on my wishlist, there are only no free resources currently. If we have a spec for configuration files, I see no problem to adjust libeconf to follow that spec.

I'll leave it optional in the spec - it is a nice feature to have for users, however it is not necessary to achieve the main goals of hermetic-usr and stateless systems

specs/configuration_files_specification.md
bluca: That sounds way too specific to me, the purpose of this is to tell people "use libeconf if you don't have anything, and if you really really have to do your own thing, follow these basic rules so that we can get a stateless system"

jluebbe (approved):

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Masking
+As a special override case, it must be possible to mask files across different
+locations by creating a symlink to `/dev/null` or an empty file.
+
+For example, an empty `/etc/foo/bar.conf` means that `/usr/foo/bar.conf` is
+masked and thus not parsed.
+
+## Drop-ins
+All configuration paths must support drop-ins. Supporting drop-ins means that in
```
behrmann, line 58: I guess one would have to wilfully misunderstand this, but maybe note later that dropins cannot have dropins of their own, i.e. `/etc/foo/bar.conf.d/baz.conf.d` is not a thing.

behrmann (approved):
LGTM, but could maybe spell out a few more special cases, because there will be people implementing this.

An example list like at the end of the version format specification that could double as a test suite, would be great.

specs/configuration_files_specification.md
```diff
@@ -0,0 +1,79 @@
...
+
+For example, settings in `/usr/foo/bar.conf.d/a.conf` must override settings in
+`/etc/foo/bar.conf`.
+
+### Drop-ins without Main Configuration File
+Optionally, schemes with only drop-ins, without a 'main' configuration file, should also
+be supported by implementations. In such schemes many drop-ins are loaded from a common
+directory in each hierarchy.
+
+For example, `/usr/foo.d/a.conf`, `/usr/foo.d/b.conf` and `/etc/foo.d/c.conf` are all
```
thkukuk, line 78: We use:
* `/usr/lib/foo/bar.conf.d`
* `/usr/foo/bar.conf.d`
* `/usr/foo.d/?.conf`
Could we please somehow unify the examples? That would make reading and understanding much easier.
I would suggest to use in all three cases `/usr/lib/foo/bar.conf.d`

thkukuk (changes requested):

thkukuk (approved):

jluebbe (approved):

pothos:
Thanks, this is much more readable than the previous try of extending the xdg spec.

PR #84, opened 2023-10-04
specs/config-files: say that drop-ins are not always required

### LINKED ISSUE #48: Drop-ins should not be mandatory
From the Base Directory Specification:

> All configuration paths must support drop-ins. Supporting drop-ins means that in addition to parsing $XDG_CONFIG_DIRS/foo/bar, an implementation also must parse all files in $XDG_CONFIG_DIRS/foo/bar.d/, and merge the result in lexicographic order. 

While drop-in directories are a good idea for many applications (e.g. for integration of multiple applications & to facilitate package upgrades), some configuration file formats can’t be merged, or (unstructured) merging might not be desirable (e.g. for security reasons, or because it adds too much complications), so the specification should not make that mandatory. Recommending that it should be implemented where possible/desirable would make more sense.

keszybz:
Also drop the sentence that "drop-ins are not full replacements" — in practice they are and I'm not sure what this sentence was trying to achieve.

Resolves #48.

### DIFF
--- a/specs/configuration_files_specification.md
+++ b/specs/configuration_files_specification.md
@@ -55,20 +55,23 @@ For example, an empty `/etc/foo/bar.conf` means that `/usr/lib/foo/bar.conf` is
 masked and thus not parsed.
 
 ## Drop-ins
-All configuration paths must support drop-ins. Supporting drop-ins means that in
-addition to parsing a full configuration file, an implementation also must parse all
-files in the drop-in directory associated with it, and merge the result in lexicographic
-order. Drop-ins are not full replacements, but must allow specifying a subset of settings
-present in the main configuration files.
+All configuration paths must support drop-ins,
+except for configuration file formats where automatic combining of multiple files is not feasible,
+for example scripts or structured documents.
+Supporting drop-ins means that in addition to parsing a full configuration file,
+an implementation also parses the drop-in files in the drop-in directories associated with it.
 
-For example, in addition to parsing `/etc/foo/bar.conf`, `/etc/foo/bar.conf.d/a.conf` and
-`/etc/foo/bar.conf.d/b.conf` must also be parsed, in this order.
+Drop-ins always have higher precedence than the configuration file they refer to.
+Drop-ins are sorted in the lexicographic order using the file name without the path,
+regardless of the hierarchy under which they are stored.
+The drop-ins that are later in this order have higher precedence.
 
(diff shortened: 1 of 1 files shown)


specs/configuration_files_specification.md
```diff
@@ -55,11 +55,13 @@ For example, an empty `/etc/foo/bar.conf` means that `/usr/lib/foo/bar.conf` is
...
 ## Drop-ins
-All configuration paths must support drop-ins. Supporting drop-ins means that in
-addition to parsing a full configuration file, an implementation also must parse all
-files in the drop-in directory associated with it, and merge the result in lexicographic
-order. Drop-ins are not full replacements, but must allow specifying a subset of settings
-present in the main configuration files.
+All configuration paths must support drop-ins.
+Supporting drop-ins means that in addition to parsing a full configuration file,
+an implementation also must parse all files in the drop-in directory associated with it,
+and merge the result in lexicographic order.
```
poettering, line 61: if you mention the order here I think you should also say that drop-ins are supposed to override relevant settings in the file they extend

poettering: also, lexicographic order is one thing, it also matters to say that settings in the drop-ins are supposed to override earlier settings in early drop-ins

bluca (approved):

specs/configuration_files_specification.md
keszybz: > if you mention the order here I think you should also say that drop-ins are supposed to override relevant settings in the file they extend

This is covered in the second paragraph below.

> also, lexicographic order is one thing, it also matters to say that settings in the drop-ins are supposed to override earlier settings in early drop-ins

I rewrote the whole section to be more explicit about the sort order. Also added an example with config files under different hierarchies, which is more interesting.

specs/configuration_files_specification.md
```diff
@@ -88,4 +91,5 @@ be supported by implementations. In such schemes many drop-ins are loaded from a
 directory in each hierarchy.
 
 For example, `/usr/lib/foo.d/a.conf`, `/usr/lib/foo.d/b.conf` and `/etc/foo.d/c.conf`
-are all loaded and parsed in this scheme.
+are all loaded and parsed in this scheme,
+in this order.
```
bluca, line 95: this line break looks a bit too eager? just a nit

keszybz: I removed the line break.

DaanDeMeyer (approved):

PR #86, opened 2023-10-09
specs: DPS: add GUIDs for big-endian MIPS architecture

rohieb:
MIPS comes in little as well as big endianess, which is determined at system boot and usually cannot be changed after startup. Binutils and GCC use the arch prefix `mips` for the big-endian variant and `mipsel` for the little-endian variant, so use the same here.


bluca (approved):
please send a PR to update https://github.com/systemd/systemd/blob/main/src/systemd/sd-gpt.h too

PR #88, opened 2023-11-16
ddi-spec: define suggested suffixes, mime types, sector sizes for DDIs

poettering:


specs/discoverable_disk_image.md
```diff
@@ -17,7 +17,18 @@ into one.  They are designed to be composable and stackable, and provide securit
...
-dm-verity signature partition's JSON content.
+`dm-verity` signature partition's JSON content.
+
+It is recommended to use a sector size of 512 bytes or 4096 for DDIs. Software operating with DDIs should
+automatically derive the sector size used for a DDI by looking for the "EFI PART" magic string at offsets 512
+or 4096, as per GPT specification.
+
+## Naming
+DDIs should use `.raw` as file suffix. A secondary suffix may be used to clarify the specific usage class of
+a DDI. For now the two secondary suffixes `.sysext.raw` and `.confext.raw` are defined (for system extension
```
bluca, line 28: maybe make the text in brackets a link to specs/extension_image.md

poettering: added a link.

bluca (approved):

keszybz:
I think the extension ".raw" is bad. It's too generic: there are simply too many things which are "raw", and different people will have different expectations, and an extension should be unique. OTOH, a DDI is not actually "raw", it has a header with a partition table, so it's a well-defined binary container format. "Raw" would normally be used for file with just contains unadorned data, without a header or other means of understanding the contents. E.g. a "raw file with measurements" could be a file with a bunch of floating-point numbers in binary format. Once you take the exact same data, but wrap a HDF5 header around it, it's not "raw".

bluca:
Afraid that ship has sailed long ago, as this is now used everywhere

keszybz:
Nah, not that widely. And actually writing this out made me realize why it extension always seemed wrong.

https://github.com/uapi-group/specifications/pull/89 is a proposal for discussion.

PR #89, opened 2023-11-17
DDI: change suffix to .gpt

keszybz:
Follow-up to 68dd1d96c359908cf494454af2b0ce1b9d4f9599: the extension ".raw" is not ideal:
- it is too generic: there are simply too many things which are "raw", and different people will have different expectations, and an extension should be unique.
- a DDI is not actually "raw". It has a header with a partition table, so it's a well-defined binary container format. "Raw" would normally be used for file with just contains unadorned data, without a header or other means of understanding the contents without context. Once you take the exact same data, but wrap a header around it that describes the data, the file contents are not "raw".

Semantic Line Breaks à la https://sembr.org/ are used.

### DIFF
--- a/specs/discoverable_disk_image.md
+++ b/specs/discoverable_disk_image.md
@@ -25,16 +25,18 @@ or 4096, as per GPT specification.
 
 ## Naming
 
-DDIs should use `.raw` as file suffix. A secondary suffix may be used to clarify the specific usage class of
-a DDI. For now the two secondary suffixes `.sysext.raw` and `.confext.raw` are defined (for system extension
-DDIs and configuration extension DDIs, see [Extension
-Images](https://uapi-group.org/specifications/specs/extension_image) for details).
+DDIs should use `.gpt` as file suffix.
+A secondary suffix may be used to clarify the specific usage class of a DDI.
+For now the two secondary suffixes `.sysext.gpt` and `.confext.gpt` are defined
+(for system extension DDIs and configuration extension DDIs, see
+[Extension Images](https://uapi-group.org/specifications/specs/extension_image)
+for details).
 
 The MIME type for DDIs is `application/vnd.efi.img`, [as per
 IANA](https://www.iana.org/assignments/media-types/application/vnd.efi.img).
 
 ## Image Version
 If the DDI is versioned, the version format described in the
-[Version Format Specification](version_format_specification.md) must be used. The underscore character (`_`)
-must be used to separate the version from the name of the image. For example: `foo_1.2.raw` denotes a `foo`
-DDI with version `1.2`.
+[Version Format Specification](version_format_specification.md) must be used.
+The underscore character (`_`) must be used to separate the version from the name of the image.
+For example: `foo_1.2.gpt` denotes a `foo` DDI with version `1.2`.
--- a/specs/extension_image.md
+++ b/specs/extension_image.md
@@ -39,9 +39,10 @@ Extension Images should be additive, and not override content present in the bas
 but this will not be enforced.
 
 ## File Suffix
-Since extensions images are DDIs, they should carry the `.raw` suffix. In order to make discerning system
-extensions and configuration extensions easy it is recommended to use the `.sysext.raw` suffix for system
-extensions, and `.confext.raw` for configuration extensions.
+Since extension images are DDIs, they should also carry the `.gpt` suffix.
+In order to make discerning system extensions and configuration extensions easy
+it is recommended to use the `.sysext.gpt` suffix for system extensions,
+and `.confext.gpt` for configuration extensions.
 
 ## Identification
 An Extension Image must contain a `extension-release.<IMAGE>` file, where `<IMAGE>` must either match the


bluca:
Nonono very hard no. I just spent ~4 years switching everything from what I had before to .raw as @poettering wanted (it's not even fully finished yet). It required an astonishing amount of work from several departments and groups. I am not going to switch again.

keszybz:
That's why such things should be discussed publicly first. When an implementation is created internally, then it's subject to change later. But anyway, I don't think you actually have to change anything internally. This is just a recommendation after all, and you can use `.raw` or anything else internally without any problem.

bluca:
It was discussed publicly, on various systemd PRs and issues. Not here obviously because this org didn't even exist at the time. But it is the standard now and it cannot change, as it's an established API.

keszybz:
> It was discussed publicly, on various systemd PRs and issues. Not here obviously because this org didn't even exist at the time. But it is the standard now and it cannot change, as it's an established API.

Of course it can be changed. The old suffix would have to be honoured, but that's no problem. (I think that in most places we don't care about the suffix at all, but there are some places where we enumerate things.)  As far as transitions go, this should be a very easy one.

bluca:
No, because then tooling would start to drop support for the existing one, or new tooling would not even add it, breaking all existing use cases, for example all image creation tools would stop using the existing suffix and create incompatible images.

poettering:
We had this discussion back at LPC in Dublin last year already, with you @keszybz. Back then we already more or less came to the conclusion that the ship has sailed on the .raw suffix, it's too much established already, and our codebase only cares for that suffix, for example the code in src/shared/discover-image.c only knows how to deal with that suffix, and so does nspawn and so on. (they'll take file wit any suffix, but only if the suffix is .raw know how to chop it off correctly to get a default hostname and such)

Our compromise back then was to introduce secondary suffixes though, i.e. `.confext.raw` and `.sysext.raw`. I think that's a nice approach. i.e. the primary suffix `.raw` indicates this is a disk image, and the secondary suffix indicates what's in it.

I added code for dealing with these secodnary suffix to discover-image.c in the past months btw. I think with v255 we should be reasonably complete on that.

the uefi forum people registered with IANA that GPT disk images should have a mime type of "application/vnd.efi.img" btw, and at the same time registered a suffix for the files: `.img`. See https://www.iana.org/assignments/media-types/application/vnd.efi.img.

`.img` is of course as bad a suffix as `.raw` is I guess.

XDG shared-mime-info (i.e. the thing file manages on gnome/kde/… use to identify file types) implements the IANA definition btw. I recently filed this to make this more compatible with our images: thahttps://gitlab.freedesktop.org/xdg/shared-mime-info/-/merge_requests/276

Anyway, I think we should stick to the `.raw` suffix, and maybe extend the secondary suffix thing. i.e. we could also add `.portable.raw` (or `.psvc.raw` or so) for portable services. and `.os.raw` (or `.machine.raw`) for bootable container/baremetal systems?

At the same time we should continue to support image names without the secondary suffixes, with only primary suffix to support multi-modal images, i.e. images that are both confext and sysext, or images that are both portable services and OS images.

poettering:
We might want to define proper mime types for our secondary suffixes btw (not for .raw, because too generic), maybe even register with iana, but I am too lazy for that bureaucracy, and just use one of those x mime types

keszybz:
Yeah, we discussed this and I regret not pushing harder against this extension. I didn't realize that this will be visible in so many places. As long as this was internal to our implementation, it didn't matter so much. But documenting this as the one and only extension makes this more visible and problematic.

> the uefi forum people registered with IANA that GPT disk images should have a mime type of "application/vnd.efi.img" btw, and at the same time registered a suffix for the files: `.img`. See https://www.iana.org/assignments/media-types/application/vnd.efi.img.

`.img` is not a great extension, because it's so generic. I think it's marginally better than `.raw`.

(FWIW, the description on the iana page is terrible: it doesn't even mention GPT anywhere.)

> Anyway, I think we should stick to the .raw suffix, and maybe extend the secondary suffix thing. i.e. we could also add .portable.raw (or .psvc.raw or so) for portable services. and .os.raw (or .machine.raw) for bootable container/baremetal systems?

OK, so if `.raw` is so widespread, it should be supported. But what about adding an alternative suffix? I think either `.gpt` or `.ddi` would be OK. E.g. `.confext.ddi` is very informative.

poettering:
So I am not convinced this is really necessary, but I am also not totally against it, if these suffixes are supported as an alternative in our codebase. But please prep the systemd commit for this at the same time, as I am really not a fan of specs that have no footing in reality.

poettering:
.gpt is not a great suffix btw, .ddi would be much better. but still, I am not convinced this all is really necessary.

sonukaloshiya:
#89

poettering:
One small update: UEFI actually cares about the file suffix a lot: it's device path logic (which becomes visible during network booting) will set up ramdisks only if the file suffix is ".img".

I am tempted to dictate that that in contexts where this matters, i.e. where UEFI shall mount a disk natively as ramdisk, DDIs should use the ".raw.img" suffix, so that it's both named a bit like our stuff, and UEFI is happy too.

bluca:
Why does this matter for UEFI boot? That's for UKIs no?

poettering:
> Why does this matter for UEFI boot? That's for UKIs no?

uefi boot does two things: booth from PE binaries and boot from GPT disk images which are loaded into a ramdisk. You can point UEFI to an URL pointing to either of these types, and it will do the right thing. The determination what to do is based on mime type first (which noone sets), and then by suffix. if the suffix is .img it will to ramdisk stuff and if it is efi it will invoke the donwloaded thing as PE binary directly.

Hence i figure in order to improve compat with the UEFI boot world we probably should always make DDI images that have a suffix of .raw normally also available under .raw.img, so that one can directly boot them with uefi http boot.

bluca:
Given it's a workaround for that specific use case, it would be fine to document a suggestion to add a symlink to that effect to cover it, but the canonical name should remain the same. We can't break backward compatibility on this

PR #91, opened 2023-12-11
UKI: document where UKI addons provided via package should be

esposem:
According with
https://github.com/systemd/systemd/issues/29372#issuecomment-1843130973

### DIFF
--- a/specs/unified_kernel_image.md
+++ b/specs/unified_kernel_image.md
@@ -159,3 +159,36 @@ that tools staging or consuming UKIs have a common place to store and look for t
 
 The installed UKIs should have a filename `<version format specification>.efi`, i.e. the filename is left to
 implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).
+
+### Auxiliary resources
+
+Auxiliary UKI resources (such as PE addons for command line extensions and similar,
+as well as systemd-sysext and systemd-confext DDIs) built centrally by distributions and
+installed via package manager should be installed into locations depending on
+whether they should be applied to all UKIs installed in the ESP, or only to a
+single specific UKI.
+
+UKI auxiliary resources that apply to *all* installed UKIs should be
+installed into `/usr/lib/modules/uki.extra.d/`. UKI auxiliary resources that
+apply to *one* specific installed UKI should be instead installed into
+`/usr/lib/modules/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of
+`uname -r` of the kernel included in the UKI and `$UKI` is the name of the
+corresponding centrally built UKI with the `.efi` extension stripped.
+
+The installed UKI auxiliary resources must have a specific file extension, which
+depends on the resource type:
+* `.addon.efi` for PE addons,
+* `.sysext.raw` for sysext DDIs,
+* `.confext.raw` for confext DDIs
+
+#### Example
+
+Given a UKI `bar_123.efi` that includes a kernel `6.9.1-1.foo`, consider
+* a PE addon `machine-id` that should apply to all installed UKIs,
+* a PE addon `proprietary-driver_2000` that is specific to the `bar_123` UKI, and
+* a sysext `mysysext_1.23.47^3` that should apply to all installed UKIs.
+
+The resulting paths would be
+* `/usr/lib/modules/uki.extra.d/machine-id.addon.efi`,
+* `/usr/lib/modules/6.9.1-1.foo/bar_123.efi.extra.d/proprietary-driver_2000.addon.efi`, and
+* `/usr/lib/modules/uki.extra.d/mysysext_1.23.47^3.sysext.raw`.


specs/unified_kernel_image.md
```diff
@@ -146,3 +146,10 @@ that tools staging or consuming UKIs have a common place to store and look for t
 
 The installed UKIs should have a filename `<version format specification>.efi`, i.e. the filename is left to
 implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).
+
+UKI addons (cmdline PE addons, systemd-sysext extensions for initrd) built centrally by distributions and
+installed via package manager should instead be installed in `/usr/lib/linux/extra.d/` if they are global
+addons/extensions, i.e. they are applied to all UKI installed in the ESP, and
+`/usr/lib/linux/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of `uname -r` of the kernel included
+in the UKI, and `$UKI` is the UKI name just as in `/boot/efi/EFI/Linux/$UKI.efi` and its addons folder
```
Foxboron, line 154: I'm pretty sure the proper ESP would be `/efi` and not `/boot/efi`?

esposem: In Fedora I don't see `/efi` for example, I think it depends on the distro

DaanDeMeyer: Because Fedora doesn't follow the Boot Loader Specification. We should use the paths from the Boot Loader Specification, not what Fedora uses here.

esposem: Updated with `$BOOT`. Please let me know if it's okay for you.

bluca (approved):

specs/unified_kernel_image.md
```diff
@@ -146,3 +146,10 @@ that tools staging or consuming UKIs have a common place to store and look for t
 
 The installed UKIs should have a filename `<version format specification>.efi`, i.e. the filename is left to
 implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).
+
+UKI addons (cmdline PE addons, systemd-sysext extensions for initrd) built centrally by distributions and
+installed via package manager should instead be installed in `/usr/lib/linux/extra.d/` if they are global
+addons/extensions, i.e. they are applied to all UKI installed in the ESP, and
+`/usr/lib/linux/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of `uname -r` of the kernel included
+in the UKI, and `$UKI` is the UKI name just as in `$BOOT/EFI/Linux/$UKI.efi` and its addons folder
+`$BOOT/EFI/Linux/$UKI.efi.extra.d/` (see [`$BOOT`](./boot_loader_specification.md#the-boot-partition-placeholder)).
```
behrmann, line 155: I have a hard time parsing this sentence. Maybe breaking this into two sentences could make this easier to understand?

So the intention is:
1. Install a global UKI addon to `/usr/lib/linux/extra.d/`.
2. Install a UKI addon for a specific UKI to `/usr/lib/linux/$UNAME/$UKI.efi.extra.d/` and `$BOOT/EFI/Linux/$UKI.efi.extra.d/`.

Is that correct? I'm sorry if it's not the multiple ands confused me a bit.

If it's correct how about something like this
> UKI addons (cmdline PE addons, systemd-sysext extensions for the initrd) built centrally by distributions and
installed via package manager should be installed into locations depending on whether they are global UKI addons, that should be applied to all UKIs installed in the ESP, or whether they should only be applied to a single specific UKI. 
>
> Global UKI addons should be installed into `/usr/lib/linux/extra.d/`. UKI-specific addons should be installed to `/usr/lib/linux/extra.d/` and `$BOOT/EFI/Linux/$UKI.efi.extra.d/`, where `$UNAME` is the output of `uname -r` of the kernel included in the UKI, `$UKI` is the UKI name just as in `$BOOT/EFI/Linux/$UKI.efi`, and `$BOOT` is as it is defined in the [Boot Loader Specification](./boot_loader_specification.md#the-boot-partition-placeholder).

behrmann (commented):
Sorry for the late review, somehow I missed the notification

specs/unified_kernel_image.md
esposem: No, I didn't want to mean that. But you are right, if it confuses, we need to write it better :smile: 
The intention is:
* global addons stay in `/usr/lib/linux/extra.d/`
* local addons stay in`/usr/lib/linux/$UNAME/$UKI.efi.extra.d/`.

What is `$UNAME`?  the output of `uname -r` of the kernel included in the UKI that is being referred by the local addon.
What is `$UKI`? It's the UKI name referred by the local addon, the same used in `$BOOT/EFI/Linux/$UKI.efi` .
What is `$BOOT`? it is defined in the [Boot Loader Specification](https://github.com/uapi-group/specifications/pull/boot_loader_specification.md#the-boot-partition-placeholder).

esposem: So it should be
```
UKI addons (cmdline PE addons, systemd-sysext extensions for the initrd) built centrally by distributions and installed via package manager should be installed into locations depending on whether they are global UKI addons, that should be applied to all UKIs installed in the ESP, or whether they should only be applied to a single specific UKI.

Global UKI addons should be installed into `/usr/lib/linux/extra.d/`. UKI-specific addons should be installed to `/usr/lib/linux/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of `uname -r` of the kernel included in the UKI, `$UKI` is the UKI name just as in $BOOT/EFI/Linux/$UKI.efi, and $BOOT is defined in the [Boot Loader Specification](https://github.com/uapi-group/specifications/pull/boot_loader_specification.md#the-boot-partition-placeholder).
```

behrmann: Yes, I think that's good. Thank you!

There remains the question of the path. There was a longer discussion in https://github.com/uapi-group/specifications/pull/60 that then settled to `/usr/lib/modules`. At least for sysexts as addons, I think the idea in that discussion was to have them go alongside. My understanding is, that you are choosing `/usr/lib/linux`, since there is prior art in Ubuntu (if I remember correctly). I can see arguments in either direction, especially for global addons, but for the UKI specific ones that is something that should be properly hashed out so that everyone is happy. 

I can see the following possibilities
- go with `/usr/lib/linux/$UNAME/$UKI.efi.extra.d/`
- go with /usr/lib/modules/$UNAME/$UKI.efi.extra.d/`
- allow both as independent directories
- allow symlinks
---
Having written all that, I have two thoughts:

1. > $UKI is the UKI name just as in $BOOT/EFI/Linux/$UKI.efi, and $BOOT is defined in the [Boot Loader Specification](https://github.com/uapi-group/specifications/pull/boot_loader_specification.md#the-boot-partition-placeholder).

I wonder if this could be made easier. We leave the name of the UKI in its place under `/usr/lib/modules` completely open, so should the maybe refer to that one instead of the one under `$BOOT`? Otherwise it is not necessarily clear what belongs together just from looking at `/usr/`.

2. Maybe this should also mention that the filenames for the addons should be valid to be compared by the version format specification as well.

bluca: having the UKI in one hierarchy and the addons in another doesn't make much sense to me, we should use the same location for both (I do not mind which one personally)

esposem: This same thought occourred to me yesterday too. I commented in https://github.com/systemd/systemd/issues/29372#issuecomment-1983260633 as we are mainly discussing it there.
 
Regarding `$UKI`, I think it's not a huge problem what exactly that is pointing. In my opinion, a smart choice would be to have `$UKI` in `/usr/lib/modules/$UKI` and in `$BOOT` to be the same, because it will be easy for the user and the tools coming up to pair the UKIs.
If we want to somehow force it through documentation is another point though.

I don't get point 2.

behrmann: I think the "too" is unnecessary, because the extension is different.
> The installed UKI addons should have a filename `<version format specification>.addon.efi`, i.e. the filename is left to implementers but must be valid for comparisons according to the Version Format Specification.

esposem:
Pushed the new text, but still we don't have agreement regarding where to put the addons, so this PR is going on draft until that is not decided. Once decided, I just need to change the path in the text.

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,16 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
 implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).
+
+UKI addons (cmdline PE addons, systemd-sysext extensions for the initrd) built centrally by distributions and
+installed via package manager should be installed into locations depending on whether they are global UKI addons,
+that should be applied to all UKIs installed in the ESP, or whether they should only be applied to a
+single specific UKI.
+
+Global UKI addons should be installed into `/usr/lib/modules/extra.d/`. UKI-specific addons should be installed
+to `/usr/lib/modules/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of `uname -r` of the kernel included
+in the UKI, `$UKI` is the UKI name just as in `/usr/lib/modules/$UKI.efi`.
```
behrmann, line 170: missing and end
```suggestion
in the UKI, and `$UKI` is the UKI name just as in `/usr/lib/modules/$UKI.efi`.
```

My understanding is, that `$UKI.efi` refers to the second paragraph in this section:
> The installed UKIs should have a filename `<version format specification>.efi`, i.e. the filename is left to
implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).

but the path for that in the first paragraph is `/usr/lib/modules/$UNAME/`, so should this all in all be this?
```suggestion
in the UKI, and `$UKI` is the UKI name just as in `/usr/lib/modules/$UNAME/$UKI.efi`.
```
Maybe this could also be written as 
```suggestion
in the UKI, `$UKI` is the name of the corresponding centrally built UKI with the `.efi` extension stripped.
```

It would probably hammer this home if at the end we added an example, e.g. for kernel version `6.9.1-1-foo` and and UKI name `bar-123` the UKI would be `/usr/lib/modules/6.9.1-1-foo/bar-123.efi` and its addons would go to `/usr/lib/modules/6.9.1-1-foo/bar-123.efi.extra.d/`, e.g. as `/usr/lib/modules/6.9.1-1-foo/bar-123.efi.extra.d/abc-456.addon.efi`. Does that sound correct?

esposem: Makes sense

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
 implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).
+
+UKI addons (cmdline PE addons, systemd-sysext extensions for the initrd) built centrally by distributions and
+installed via package manager should be installed into locations depending on whether they are global UKI addons,
+that should be applied to all UKIs installed in the ESP, or whether they should only be applied to a
+single specific UKI.
+
+Global UKI addons should be installed into `/usr/lib/modules/$UNAME/extra.d/`. UKI-specific addons should be installed
+to `/usr/lib/modules/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of `uname -r` of the kernel included
+in the UKI and `$UKI` is the name of the corresponding centrally built UKI with the `.efi` extension stripped.
```
keszybz, line 170: Hmm. The first part of the text says "global UKI addons, that should be applied to all UKIs installed in the ESP". When I read that, I expect a path that doesn't refer to `$UNAME`. I would expect a path like `/usr/lib/modules/extra.d`. This would apply to any addon which doesn't depend strongly on the kernel version.

esposem: Keeping in mind that anything is fine for me at this point, I also initially thought of having global add-ons in modules/extra.d/. But then think about multiple rpms of multiple kernel version, each installing its own global addons. Maybe for kernel X the global addon test.addon.efi has test=1, for kernel Y the same addon has test=2?
Does it make sense to you? Or would that not be a problem?

keszybz: Well, if you have multiple rpms that want to provide global addons with the same name, that clearly wouldn't work. A global addon would need to be packaged separately.

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+
+### Example
+Given `UNAME="6.9.1-1-foo"` and `UKI="bar.efi"`:
+* UKI will be in `/usr/lib/modules/6.9.1-1-foo/bar.efi`
+* Global addons will be in `/usr/lib/modules/6.9.1-1-foo/extra.d/`
+* Local addons will be in `/usr/lib/modules/6.9.1-1-foo/bar.efi.extra.d/`
```
behrmann, line 176: Could you move the example to the bottom? Otherwise the following paragraph is part of the example section.

esposem:
So to sum up:
Global go to `/usr/lib/modules/extra.d/` and local addons go in `/usr/lib/modules/<uname-r>/<uki-name>.efi.extra.d/`.

bluca:
modules is a directory owned by kernel packages, adding an extra.d is too generic, the name needs to be more specific

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+
+The installed UKI addons should have a filename `<version format specification>.addon.efi`, i.e. the filename
+is left to implementers but must be valid for comparisons according to the
+[Version Format Specification](version_format_specification.md).
+
+### Example
+Given `UNAME="6.9.1-1-foo"` and `UKI="bar.efi"`:
+* UKI will be in `/usr/lib/modules/6.9.1-1-foo/bar.efi`
+* Global addons will be in `/usr/lib/modules/extra.d/`
+* Local addons will be in `/usr/lib/modules/6.9.1-1-foo/bar.efi.extra.d/`
```
aplanas, line 180: Sorry if I am asking an already answered question!  But is there a place for user defined global and local addons?

For example, if an user want to add a new cmdline PE addons, that for example enable some debug output? I am also thinking on immutable systems, where the user cannot (should not) change the rootfs content (outside the package manager).

esposem:
> modules is a directory owned by kernel packages, adding an extra.d is too generic, the name needs to be more specific

I'll go with `uki.extra.d` then

specs/unified_kernel_image.md
esposem: The idea is to have all rootfs addons in these two folders as agreed location. You can put your addons anywhere you want in the root fs, they will be considered only if they are present in the ESP.
So if  the user wants to create its own debug output, and the system is immutable, the addon can stay anywhere he wants, as long as it is then copied in the ESP when needed.

behrmann:
Looks good to me, but
> I'll go with `uki.extra.d` then

seems to still be missing?

esposem:
Apologies. I was sure I pushed :(

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
 
 The installed UKIs should have a filename `<version format specification>.efi`, i.e. the filename is left to
 implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).
+
+UKI addons (cmdline PE addons, systemd-sysext extensions for the initrd) built centrally by distributions and
+installed via package manager should be installed into locations depending on whether they are global UKI addons,
+that should be applied to all UKIs installed in the ESP, or whether they should only be applied to a
+single specific UKI.
+
+Global UKI addons should be installed into `/usr/lib/modules/uki.extra.d/`. UKI-specific addons should be installed
```
poettering, line 168: I'd not use the word "Global" here, since it might not be clear what it means.

hence maybe:

"UKI addons that shall apply to *all* installed kernels may be installed into `/usr/lib/modules/uki.extra.d/`. UKI addons that shall apply to a specific installed kernel may be installed into  `/usr/lib/modules/$UNAME/$UKI.efi.extra.d/`, where …"

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+to `/usr/lib/modules/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of `uname -r` of the kernel included
+in the UKI and `$UKI` is the name of the corresponding centrally built UKI with the `.efi` extension stripped.
+
+The installed UKI addons should have a filename `<version format specification>.addon.efi`, i.e. the filename
```
poettering, line 172: this might be misread as if addons should not carry identifying names, just version numbers. But I'd assume the opposite: that they more likely have identifying names, and less likely version ids.

i.e. let's say we start passing the machine ID as an addon into the booted system (so that the initrd can start using it, before transitioning into the host). In that case the addon should probably just be called "machine-id.addon.efi", without any version, since the version is never going to change.

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+is left to implementers but must be valid for comparisons according to the
+[Version Format Specification](version_format_specification.md).
+
+### Example
+Given `UNAME="6.9.1-1-foo"` and `UKI="bar.efi"`:
+* UKI will be in `/usr/lib/modules/6.9.1-1-foo/bar.efi`
+* Global addons will be in `/usr/lib/modules/uki.extra.d/`
+* Local addons will be in `/usr/lib/modules/6.9.1-1-foo/bar.efi.extra.d/`
```
poettering, line 180: I think the example would be a lot more helpful if it'd would actually mention a specific versioned addon file name.

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
 
 The installed UKIs should have a filename `<version format specification>.efi`, i.e. the filename is left to
 implementers but must be valid for comparisons according to the [Version Format Specification](version_format_specification.md).
+
+UKI addons (cmdline PE addons, systemd-sysext extensions for the initrd) built centrally by distributions and
```
poettering, line 163: hmm, please don't call systemd-sysext stuff "addons".

addons → those efi PE wrapped things
sysexts → dm-verity enabled GPT disk images (DDIs)

also, we should cover confext here too, it's a thing now. And a pretty useful one at that.

Hence maybe say:

> Auxiliary UKI resources (such as PE addons for command line extensions and similar, as well as systemd-sysext and systemd-confext DDIs) …

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,22 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+installed via package manager should be installed into locations depending on whether they are global UKI addons,
```
poettering, line 164: i'd drop the "global", it's redundant, and it might mean different things to different people

poettering: and the word "addons" I'd drop too, for the aforementioned reasons (here and everywhere). Just call them "auxiliary resources" or so.

poettering:
So the chosen dirs make sense to me, I am not yet in love with the text though. i.e. it starts out with covering EFI addons *and* sysexts and then uses the word "addons" for both, even though we (at least from a systemd pov) just use the word "addon" for the EFI PE wrapped thingies, and treat "sysext" and "confext" as something separated (since they are GPT wrapped thingies).

Hence, I think we should make clear initially that this spec covers all three, and then should continue to generically talk about them as "auxiliary resources" or some generic term like that.

but the spec should document the right suffix for each:

* addon: `*.addon.efi`
* sysext: `*.sysext.raw`
* confext: `*.confext.raw`

specs/unified_kernel_image.md
esposem: check if now it's better

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,41 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+shall apply to *one* specific installed kernel shall be instead installed into
+`/usr/lib/modules/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of
+`uname -r` of the kernel included in the UKI and `$UKI` is the name of the
+corresponding centrally built UKI with the `.efi` extension stripped.
+
+The installed UKI auxiliary resources must have a specific file extension, which
+depends on the resource type:
+* for PE addons is `*.addon.efi`
+* for sysext DDIs is `*.sysext.raw`
+* for confext DDIs is `*.confext.raw`
```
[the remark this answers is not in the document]
behrmann: The paragraph is now missing any reference to the version format spec. While I agree that we need to find a wording, that cannot be misunderstood as this needing to be a versioned file name and can be something like `machineid` (though that's valid under the version format spec), we need to be clear, that the name must be valid under the spec, e.g. not use characters disallowed by it.

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,41 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+
+
+### Example
+Given the following parameters:
+* `ARCH=x86_64`
+* `DISTRO=fedora`
+* `UKI="bar.efi"`
+* `UKI_KERNEL_VERSION="6.9.1-1-foo"`
+* `RESOURCE_NAME="test"`
+* The resource type is `PE addon`, meaning the file extension is `.addon.efi`
```
behrmann, line 192: Since the extension follows from the type, the part after the comma would be better in the second paragraph, but I'd just drop it, since it appears there already and it follows from the text and the parameters given.

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,41 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+
+This is where an UKI/auxiliary resources vendor installs the files:
+* Auxiliary resource filename is `test-bar.fedora.x86_64.addon.efi`
```
behrmann, line 195: This and the `ARCH=` and `DISTRO=` above confuse me a bit. This example is the first were this comes into play and my understanding is, that e.g. `machine-id.addon.efi` would be valid. Having the example like this would probably be easily misread, that the template should always have a form `NAME.DISTRO.ARCH.TYPE` or that `NAME` in there must refer to the name of the UKI again.

How about something like this?
> Given a UKI `bar-123.efi` that includes a kernel with the release information `6.9.1-1-foo`. Consider
> - a PE addons `machine-id` that should apply to all installed UKIs,
> - a PE addons `proprietary-driver-2000` that is specific to one installed UKI, and
> - a sysext `mysysext-1.23.47^3` that again should apply to all installed UKIs.
>
> The resulting paths would be
> - `/usr/lib/modules/uki.extra.d/machine-id.addon.efi`,
> - `usr/lib/modules/6.9.1-1-foo/bar-123.efi.extra.d/proprietary-driver-2000.addon.efi`, and
> - `/usr/lib/modules/uki.extra.d/mysysext-1.23.47^3.sysext.raw`.

to cover most bases?

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,41 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+
+### Auxiliary resources
+
+Auxiliary UKI resources (such as PE addons for command line extensions and similar,
+as well as systemd-sysext and systemd-confext DDIs) built centrally by distributions and
+installed via package manager should be installed into locations depending on
+whether they should be applied to all UKIs installed in the ESP, or only to a
+single specific UKI.
+
+UKI auxiliary resources that shall apply to *all* installed kernels may be
```
behrmann, line 171: Should this maybe be
```suggestion
UKI auxiliary resources that shall apply to *all* installed UKIs may be
```
since for a given kernel package multiple or no UKIs could be installed.

specs/unified_kernel_image.md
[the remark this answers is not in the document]
esposem: I am afraid that I wouldn't know what to write here. I personally think that the sentence I had initially was good, just as in the UKI name, so if you have any proposal feel free to write it here.

specs/unified_kernel_image.md
esposem: ARCH and DISTRO are actually used in a concrete implementation of this, see https://gitlab.com/cki-project/kernel-ark/-/merge_requests/2917. When @poettering told me to use "specific versioned addon file name" I thought he meant that. But I agree that `machine-id.addon.efi` is valid, the versioning proposed there is just what we are using internally when deploying an RPM, it doesn't have to apply everywhere.

esposem: I think I will just use your proposal

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,36 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+corresponding centrally built UKI with the `.efi` extension stripped.
+
+The installed UKI auxiliary resources must have a specific file extension, which
+depends on the resource type:
+* `*.addon.efi` for PE addons,
+* `*.sysext.raw` for sysext DDIs,
+* `*.confext.raw` for confext DDIs
+
+
+### Example
```
behrmann, line 185: Nit: double empty lines above the headline, but none under it. Also, since the example should probably be part of the *Auxiliary resources* section, it needs to go one level deeper 
```suggestion
#### Example

```

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,36 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+shall apply to *one* specific installed kernel may be instead installed into
+`/usr/lib/modules/$UNAME/$UKI.efi.extra.d/`, where `$UNAME` is the output of
+`uname -r` of the kernel included in the UKI and `$UKI` is the name of the
+corresponding centrally built UKI with the `.efi` extension stripped.
+
+The installed UKI auxiliary resources must have a specific file extension, which
+depends on the resource type:
+* `*.addon.efi` for PE addons,
+* `*.sysext.raw` for sysext DDIs,
+* `*.confext.raw` for confext DDIs
```
behrmann, line 182: Not removing the `*` was silly of my suggestion, since this talks about the extensions now, they are no longer needed
```suggestion
* `.addon.efi` for PE addons,
* `.sysext.raw` for sysext DDIs,
* `.confext.raw` for confext DDIs
```

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,36 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+
+Auxiliary UKI resources (such as PE addons for command line extensions and similar,
+as well as systemd-sysext and systemd-confext DDIs) built centrally by distributions and
+installed via package manager should be installed into locations depending on
+whether they should be applied to all UKIs installed in the ESP, or only to a
+single specific UKI.
+
+UKI auxiliary resources that shall apply to *all* installed UKIs may be
+installed into `/usr/lib/modules/uki.extra.d/`. UKI auxiliary resources that
+shall apply to *one* specific installed kernel may be instead installed into
```
behrmann, line 173:
```suggestion
shall apply to *one* specific installed UKI may be instead installed into
```
missed that one earlier

behrmann, line 173: I'm not entirely sure about the requirement level, but *may* is weaker that *should*. Should this instead be
```suggestion
UKI auxiliary resources that shall apply to *all* installed UKIs should be
installed into `/usr/lib/modules/uki.extra.d/`. UKI auxiliary resources that
shall apply to *one* specific installed kernel should be instead installed into
```
since *may* is an optional thing, but this should probably be stronger?

To avoid confusion, I'd probably also just drop the "shall" in front of "apply" here.

specs/unified_kernel_image.md
```diff
@@ -159,3 +159,36 @@ that tools staging or consuming UKIs have a common place to store and look for t
...
+### Example
+Given a UKI `bar-123.efi` that includes a kernel `6.9.1-1-foo`, consider
+* a PE addon `machine-id` that should apply to all installed UKIs,
+* a PE addon `proprietary-driver-2000` that is specific to the `bar-123` UKI, and
+* a sysext `mysysext-1.23.47^3` that should apply to all installed UKIs.
+
+The resulting paths would be
+* `/usr/lib/modules/uki.extra.d/machine-id.addon.efi`,
+* `/usr/lib/modules/6.9.1-1-foo/bar-123.efi.extra.d/proprietary-driver-2000.addon.efi`, and
+* `/usr/lib/modules/uki.extra.d/mysysext-1.23.47^3.sysext.raw`.
```
poettering, line 194: we generally want people to use underscores to separate the name and the version, because "-" is kinda important for versioning.

See comments here:

https://uapi-group.org/specifications/specs/version_format_specification/

or at the bottom here:

https://uapi-group.org/specifications/specs/discoverable_disk_image/#image-version

We probably should suggest or even mandate that for versioned addons, too

keszybz (approved):
We've bikeshedded this to a gleaming polish ;) Can't do no more. Let's merge.

PR #106, opened 2024-06-22
Update version_format_specification.md

cherio:
Fixing erroneous description. By definition "Caret (^) starts a suffix that always sorts higher" however the loop logic below describes the opposite.


behrmann:
That got changed in https://github.com/uapi-group/specifications/commit/4951ead0d05b0beaa095d165147c532432cd5647 see the rationale there.

sonukaloshiya (approved):

PR #108, opened 2024-07-10
Introduce Multi-Profile UKIs

poettering:
Here's the specification for the logic implemented by this:

https://github.com/systemd/systemd/pull/33512

This should probably be merged at roughly the same time.


Itxaka:
Looking good, very clear :+1:

bluca (approved):

PR #110, opened 2024-07-16
DPS: add GUIDs for missing Debian architectures

bluca:


poettering:
x32 is not a cpu architecture, but just a different userspace abi for x86-64. I am not convinced it should be treated as architecture here. We don't have different ids for the myriad of arm abis here either.

poettering:
(the other ones i am fine with)

keszybz:
Let's have the discussion in the systemd PR and follow here with the decision made there.

keszybz:
I closed the systemd PR, so I'll all close this one. It seems we need a different approach. The discussion is continuing.

PR #125, opened 2024-10-25
Fixed a typo in what order drop-ins are being parsed.

ipanova:


keszybz (commented):
No. The point of the original text is that `b.conf` is after `a.conf`, even when one is under `/etc` and the other one under `/usr/lib`.

ipanova:
@keszybz  These specs contradict to what is being written. It says /etc takes precedence over /usr so tell me how this is right then?
`For example, in addition to parsing `/etc/foo/bar.conf`,
`/etc/foo/bar.conf.d/a.conf` and `/etc/foo/bar.conf.d/b.conf` must also be parsed,
in this order.`

keszybz:
"/etc over /usr" is the general principle, but there are various extensions/overrides. Drop-ins as a whole are used to override other configuration, so drop-ins are ordered after the main configuration file, no matter where it resides. The paragraph about dropin ordering describes how they are ordered, and this example matches that.

ipanova:
@keszybz  You are saying that if i have also have these configs `/usr/lib/foo/bar.conf` and /usr/lib/bar.conf.d/a.conf
It will go in the following order?

/usr/lib/foo/bar.conf
/etc/foo/bar.conf
/usr/lib/bar.conf.d/a.conf
/etc/foo/bar.conf.d/a.conf
/usr/lib/foo/bar.conf.d/b.conf

keszybz:
Not exactly. The files with the same name override each other. So the order is as you listed, but the crossed out files are ignored:

~~/usr/lib/foo/bar.conf~~
/etc/foo/bar.conf
~~/usr/lib/foo/bar.conf.d/a.conf~~
/etc/foo/bar.conf.d/a.conf
/usr/lib/foo/bar.conf.d/b.conf

ipanova:
If you find it useful I could add this example ( since it's more versatile) to the docs in the `drop-ins` section. Lmk.

keszybz:
Yes, please do.

PR #126, opened 2024-10-25
Add a more robust example of how config files are parsed.

ipanova:
configuration_files_specification.md: show how configs are evaluated
from different drop-in dirs in combination with colliding names and also
having separate config files.


ipanova:
@keszybz let me now if the wording is ok.

specs/configuration_files_specification.md
```diff
@@ -73,6 +73,17 @@ The configuration in `bar.conf` has the lowest priority,
 and is overridden by `a.conf` and `b.conf`.
 `b.conf` has the highest priority.
 
+Here's another, more robust example. This is the order in which the files are parsed.
```
keszybz, line 76: I don't like this introductory sentence. This is essentially the same example, but with more details. Can you combine the two paragraphs into a single description? Maybe something like: "Considering the following files present in the file system … (and then most of the text of the first paragraph)".

specs/configuration_files_specification.md
```diff
@@ -66,12 +66,18 @@ Drop-ins are sorted in the lexicographic order using the file name without the p
...
+Considering the following files are present on the filesystem, this would be the order in which the
+files are parsed. Note, that files with the same name override each other. The configuration in
+`bar.conf` has the lowest priority, and is overridden by `a.conf` and `b.conf`.`b.conf` has the
+highest priority.
+
+```
+/usr/lib/foo/bar.conf
+/etc/foo/bar.conf (overrides /usr/lib/foo/bar.conf)
+/usr/lib/foo/bar.conf.d/a.conf
+/etc/foo/bar.conf.d/a.conf (/usr/lib/foo/bar.conf.d/a.conf)
```
behrmann, line 78: "overrides" in the parentheses, just like above, would be a bit nicer.

keszybz: https://github.com/uapi-group/specifications/pull/127

ipanova: thanks, i just missed that

PR #131, opened 2024-11-12
uki: describe .efifw section

ani-sinha:
Add description for the firmware section (`.efifw`) in the specs.

See also https://github.com/systemd/systemd/pull/35091

### DIFF
--- a/specs/unified_kernel_image.md
+++ b/specs/unified_kernel_image.md
@@ -62,6 +62,7 @@ UKIs consist of the following resources:
 * Optionally, a splash image to bring to screen before transitioning into the Linux kernel, in the `.splash` section.
 * Optionally, a compiled Devicetree, for systems which need it, in the `.dtb` section.
 * Optionally, one or more compiled Devicetrees, for systems which need it, each in a separate `.dtbauto` section. The first `.dtbauto` section that matches the current hardware (matching is done either by the first `compatible` property with one from the firmware-provided Devicetree or by the SMBIOS fields using the contents of `.hwids` section as described below) will override the `.dtb` section.
+* Optionally, a firmware image in the `.efifw` section. This UEFI firmware image is an opaque binary blob that can be used as firmware in the next reset. The first `.efifw` section that matches the current hardware (matching is done first by the SMBIOS fields using the contents of `.hwids` section as described below and then by the firmware id of that matched entry) will be loaded.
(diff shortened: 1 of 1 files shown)


Foxboron:
My general feeling is that this is too generic? Is there no expectations to the embedded blob? Is this all an implementation detail?

And if it only an implementation detail, how often would this *actually* be used by others beyond the usecase in the systemd PR?

keszybz:
I think being generic here is fine. What matters is that the firmware is in some format acceptable to the hardware. We can't really say much about it this point. If this becomes established, we can add additional details later.

poettering:
Sorry for dropping the ball on this. Could you please rebase this? We really should get this landed.

Please follow the wording of the .dtbauto stuff regarding the matching.

katexochen:
I've rebased and updated this in https://github.com/uapi-group/specifications/pull/216, PTAL

ani-sinha:
> Sorry for dropping the ball on this. Could you please rebase this? We really should get this landed.
> 
> Please follow the wording of the .dtbauto stuff regarding the matching.

Sorry I too forgot about this. Let me try and get to this in the next few days.

ani-sinha:
> Sorry for dropping the ball on this. Could you please rebase this? We really should get this landed.
> 
> Please follow the wording of the .dtbauto stuff regarding the matching.

@poettering So I was trying to update the doc and looked at the code. I see that chid_match() is only called for `DEVICE_TYPE_DEVICETREE` types of device entries and not for `DEVICE_TYPE_UEFI_FW` types of entries. So it won't match against the hardware chid. I think there is a gap in the implementation.

keszybz:
We now have two versions of this PR. Since this one was earlier, I'll merge it. If subsequent changes should be done, let's do that later.

ani-sinha:
> > Sorry for dropping the ball on this. Could you please rebase this? We really should get this landed.
> > Please follow the wording of the .dtbauto stuff regarding the matching.
> 
> @poettering So I was trying to update the doc and looked at the code. I see that chid_match() is only called for `DEVICE_TYPE_DEVICETREE` types of device entries and not for `DEVICE_TYPE_UEFI_FW` types of entries. So it won't match against the hardware chid. I think there is a gap in the implementation.

I have started a PR here: https://github.com/systemd/systemd/pull/42570 Still in draft, I want to see if I introduced any regressions or not.

Issue #115, opened 2024-08-24
File Hierarchy for the Verification of OS Artifacts


dvzrv:
[OpenPGP](https://openpgp.org) is widely adopted by most Linux distributions, as it offers a decentralized system for artifact verification.

To make use of OpenPGP, most distributions establish a PGPKI (aka. [Web of Trust (WoT)](https://openpgp.dev/book/signing_components.html#wot)), in which trust is [delegated](https://openpgp.dev/book/glossary.html#term-Delegation) from a [signature](https://openpgp.dev/book/glossary.html#term-Data-Signature), over the signer's certificate to the [certificates](https://openpgp.dev/book/certificates.html) of a few central entities, that serve as [trust anchors](https://openpgp.dev/book/glossary.html#term-Trust-Anchor). This is mostly established via ["third-party certifications"](https://openpgp.dev/book/glossary.html#term-Certification).
The collection of certificates used for data signatures (sometimes in distinct roles) is distributed by many Linux distributions in bundled form, as OpenPGP keyrings.

As of writing, no standardized location for OpenPGP keyrings exists, leaving tools to either do guesswork ([related mkosi issue](https://github.com/systemd/mkosi/issues/757)) or rely on proprietary and *stateful* keystore formats (e.g. GnuPG) and a per-application location.

I am currently working towards a simple specification, that encodes metadata required by programs such as package managers, or general purpose verification tools (e.g. "which distribution?", "in what version?", "for which purpose?", "who is certifying?") in a common directory structure.
This would offer a simple, *stateless* and standardized interface allowing distributions (or anyone really) to verify their own artifacts or those of others.

bluca:
This requires buy-in from the main package managers in order to be useful - let's say apt, dnf, zypper, pacman?

dvzrv:
> This requires buy-in from the main package managers in order to be useful - let's say apt, dnf, zypper, pacman?

Yes, that would be the *ideal* outcome for the package management use-case.

My idea is to design this flexible enough, so that it can be used also for other purposes: The verification of installation media, virtual machine images, etc. (basically any artifact a distribution may produce and issue an OpenPGP signature for).

dvzrv:
While working on #134 we have tried to incorporate semantics for Sigstore as well.
However, even after consulting with experts it was not clear whether the technology could be properly modeled in the given hierarchy.
This is in large part due to the complexity of the what sigstore covers, but also due to our missing expertise and understanding of the technology.
[...]

PR #134, opened 2025-01-29
File Hierarchy for the Verification of OS Artifacts (VOA)

dvzrv:
Define a common scheme for cryptographic validation of OS artifacts. Specify a stateless directory hierarchy for the use with various cryptographic technologies, such as OpenPGP and X.509. Provide examples for the use of VOA based on verification with and without trust anchors.

Fixes: https://github.com/uapi-group/specifications/issues/115

cc @xsuchy @dkg @santiagotorres

### DIFF
--- a/specs/file_hierarchy_for_the_verification_of_os_artifacts.md
+++ b/specs/file_hierarchy_for_the_verification_of_os_artifacts.md
@@ -1,0 +1,556 @@
+---
+title: File Hierarchy for the Verification of OS Artifacts (VOA)
+category: Concepts
+layout: default
+version: 1
+SPDX-License-Identifier: CC-BY-4.0
+---
+
+# File Hierarchy for the Verification of OS Artifacts (VOA)
+
+## Motivation
+
+Cryptographic validation of artifacts with the help of digital signatures is a use-case of most Linux distributions.
+Different cryptographic technologies exist and can be used for this purpose.
+Currently, OpenPGP and X.509 are widely adopted.
+
+As of this writing, no technology-agnostic, standardized location for the distribution of cryptograpic material that serves as verifier for digital signatures exists.
+This leaves consumers to either do guesswork, or rely on proprietary, stateful or technology-specific keystore formats and per-application locations.
+
+The _File Hierarchy for the Verification of OS Artifacts (VOA)_ defines a generic approach for storage and retrieval of _signature verifiers_, supporting a wide range of cryptographic technologies.
+
+VOA aims to cover all verification needs around OS artifacts.
+
+Unlike e.g. Mozilla's Network Security Services ([NSS]), which concerns itself with secure network communication, including a curated set of X.509 CA certificates (e.g. often provided on Linux distributions via `/etc/pki/` or `/etc/ca-certificates/`), VOA focusses on the verification of artifacts with support for a diverse set of cryptographic technologies.
+
+## Terminology
+
+This text uses the generic term _"signature verifier"_ for cryptographic objects that are used to verify signatures.
+In practice, those may be for example X.509 or OpenPGP certificates (the latter are also referred to as public keys).
+
+All technologies for validation of signatures rely on _public key material_.
+However, some systems use bare _public key material_, while others combine key material with additional metadata, forming composite objects that may e.g. also make identity assertions.
+
+This text uses the term _signature verifier_ to refer to two types of verifiers:
+
+- _artifact verifiers_, which are used for the validation of signatures on artifacts,
+- _[trust anchor]s_, which are used to ascertain the authenticity of _artifact verifiers_.
+
+### Revocation of verifiers
+
+Some technologies have a concept of invalidation of verifiers, using a mechanism that signals to users that a verifier should not be relied on anymore (e.g. because the private key material has been compromised).
+
+### Classification of signature verification models
+
+For the use in the VOA structure, digital signature technologies may be regarded outside of their usual context.
+To simplify and allow a unified view on these technologies, VOA classifies these three distinct families:
+
+#### Point to Point
+
+(e.g. SSH, minisign, signify)
+
+Keys are "atomic" and don't have formal relationships with each other: A key issues a number of signatures, which can be verified.
+
+These systems use "set-style trust": Users choose to rely on some _set_ of keys/verifiers, and are willing to accept signatures made with those.
+
+This model doesn't require or support the use of _trust anchors_ in VOA.
+In this model, the set of _artifact verifiers_ is considered implicitly trusted and used directly for the verification of artifacts.
+
+#### Decentralized delegation
+
+(e.g. OpenPGP)
+
+In these systems, key material and certifications (as used in the OpenPGP "Web of Trust") are decentralized, forming a non-hierarchical network.
+Users need to specify the _trust anchors_ they want to rely on, within their context.
+
+Two approaches are possible for applications using VOA:
+
+- Directly rely on _artifact verifiers_ for a [purpose].
+- Rely on _trust anchors_ which delegate to one or more _artifact verifiers_.
+  Anchors must be explicitly defined in the relevant "trust-anchor" VOA directory (see [purpose]).
+
+#### Hierarchical delegation
+
+(e.g. X.509, SSH+CA)
+
+Other systems have an inherent assumption of hierarchical delegation, formalized by cryptographic certifications.
+In such systems, globally or locally accepted _trust anchors_ are used as starting points in signature validation.
+Certificates and the delegation relations between them in practice often form a tree structure.
+Such hierarchical key structures are often referred to as [public key infrastructure] (PKI).
+In many X.509 ecosystems a shared set of _trust anchors_ exist, that all actors agree upon (e.g. _Web PKI_).
+
+This model requires explicitly defined _trust anchors_ in VOA, that delegate to _artifact verifiers_.
+
+### Application of signature verification models
+
+Although the [classification of signature verification models] categorizes various technologies for the verification of digital signatures, the reality is more diverse.
+In practice VOA only distinguishes between direct verifications and those relying on some form of delegation.
+The details of delegation are technology specific and described in their respective [technology] sections.
+
+### Typical distribution format of verifers / short vs. long-lived keys
+
+VOA's goal is to provide a uniform representation of verifiers, to the degree that this is possible with differing technologies.
+To achieve this, verifiers are stored in a central file hierarchy that can be used with different technologies.
+
+Some concerns may be handled differently in supported technologies:
+
+- The separation between _trust anchors_ and _artifact verifiers_ may not be reasonable to model in separated directories, because the native representation of a given technology is to store both in a shared file, as a type of "certificate chain".
+- By default, VOA suggests that each verifier is stored in an individual file (that ideally reflects a unique identifier of that verifier in its filename).
+  However, in some technologies, separating verifiers into individual files may be uncommon or unreasonable.
+  In these technologies, bundles of verifiers may be stored in a shared file (this can apply equally to _artifact verifiers_ and _trust anchors_).
+- As an edge case, in some technologies, _artifact verifiers_ may be extremely short lived, so it may not be reasonable to store them in VOA.
+  In these cases, only a set of _trust anchors_ may be stored in VOA, while the _artifact verifiers_ are distributed with the signed artifacts.
+
+## File Hierarchy
+
+The _File Hierarchy for the Verification of OS Artifacts (VOA)_ is maintained as a directory structure on a system.
+It contains cryptographic public-key material for one or more technologies.
+
+The VOA hierarchy organizes _signature verifiers_ by [os], [purpose], [context] and [technology], and stores them in a directory structure of the form `$os/$purpose/$context/$technology/`.
+
+Vendors provide a set of _signature verifier_ files for their OS, in the VOA hierarchy format.
+
+Users of a VOA hierarchy (such as package installation software) can pick the relevant set of _signature verifiers_ for the verification of a specific artifact based on their location in the hierarchy.
+
+VOA does not concern itself with the retrieval of additional _signature verifiers_ or the update of existing ones in the file hierarchy.
+
+### Load paths
+
+VOA defines a list of _load paths_ with descending priority for system mode and user mode.
+
+_Signature verifiers_ are retrieved from hierarchies in these directories to validate artifacts based on specific [load logic].
+This is referred to as _verifier lookup_.
+
+The existence of multiple load paths allows system administrators and users to define custom sets of _signature verifiers_ for their system and applications.
+
+#### System mode
+
+The priority of load paths in system mode follows the definition of ["Storage Directories and Overrides" in the Configuration Files Sepcification], which is also reflected in the following list of paths:
+
+- `/etc/voa/`
+- `/run/voa/`
+- `/usr/local/share/voa/`
+- `/usr/share/voa/`
+
+#### User mode
+
+The priority of load paths in user mode mirrors that of the system mode.
+Refer to the [XDG Base Directory Specification] for details on default values.
+
+- `$XDG_CONFIG_HOME/voa/`
+- the `./voa/` directory in each directory defined in `$XDG_CONFIG_DIRS`
+- `$XDG_RUNTIME_DIR/voa/`
+- `$XDG_DATA_HOME/voa/`
+- the `./voa/` directory in each directory defined in `$XDG_DATA_DIRS`
+
+### Symlinking
+
+Load paths are constrained to self-contained locations on a host as they provide vital data for the integrity and verification of all components on a system.
+However, relative symlinks can be used in the VOA hierarchy to point to files or directories below the same [load path] or one of the _load paths_ in descending priority.
+
+As an example, symlinks to files below the same load path or to another load path with lower priority may be used to deduplicate the use of a single _signature verifier_ for multiple use-cases.
+As another example, using symlinks allows to automatically keep _signature verifiers_ in sync with canonical upstream data.
+
+VOA implementations must not consider symlinks under the following conditions and raise a warning for them instead:
+
+- the symlink is for a file or directory in an ephemeral load path (i.e. `/run/voa/` and `$XDG_RUNTIME_DIR/voa/`),
+- the symlink is for a file or directory outside of the specified _load paths_,
+- the symlink is broken (aka. "dangling"),
+- the file name of the symlink (also those of intermediate symlinks in the case of a symlink chain) does not match that of the target,
+- or the file type of the source and target of the fully resolved symlink do not match (e.g. a verifier file points to a directory, or a directory to a verifier file).
+
+### Masking
+
+Individual _signature verifiers_ may be masked using a symlink to `/dev/null`, independent of [technology].
+This constitutes the only allowed exception to the rule of not symlinking to files or directories external to the load paths.
+These symlinks are only expected to be used in writable load paths (that is, only below `/etc/voa/` or `/run/voa/` when in system mode, or below `$XDG_CONFIG_HOME/voa/` or `$XDG_RUNTIME_DIR/voa/` when in user mode).
+A _signature verifier_ that is masked is considered invalid in all [load paths] independent of technology and [load logic].
+
+Masking _signature verifiers_ may be desirable in situations where a verifier is considered compromised, but a revocation for it can not be provided to users in time.
+VOA offers this mechanism as an alternative to revocation.
+
+Masking is explicitly designed to be used on individual _signature verifiers_.
+The masking (of parts) of the directory structures that contain them on the other hand does not yield meaningful or intelligible results for end-users and would be very complex to reason about.
+VOA implementations must not consider masking symlinks for directories and should raise a warning if such symlinks are encountered.
+
+### Load logic
+
+Load logic depends on the [technology] in use.
+It is the mechanism by which relevant verifier information from all load paths is retrieved, while considering the [os], [purpose], [context] and [technology] request of a calling application.
+
+For some technologies, the first suitable _signature verifier_ found in the list of [load paths] is used, effectively [overriding] files with the same name in directories lower in the list.
+
+However, in other technologies, more complex [merging] logic may apply, that effectively combines information from different representations of one verifier.
+
+Additionally, the concept of [masking] applies for all technologies.
+
+#### Overriding
+
+In technologies that use overriding, verifier files shadow other variants of the same verifier based on filename equality.
+
+For example, `/etc/voa/fedora/packages/default/ssh/b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c.pub` overrides `/usr/share/voa/fedora/package/default/ssh/b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c.pub`.
+
+Overriding depends on file name equality and directory structure.
+
+For example, `/etc/voa/fedora/packages/default/ssh/7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730.pub` does not override `/usr/share/voa/fedora/package/default/ssh/b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c.pub`, but instead adds an additional _signature verifier_.
+
+Great care has to be taken to ensure the naming scheme has the desired overriding behavior (e.g. to make sure that a revoked version of an SSH public key overrides a non-revoked version of the same SSH public key found in a different load path).
+
+#### Merging
+
+Some technologies use a business logic that "merges" different versions of the same _signature verifier_ (instead of applying overriding semantics).
+The purpose is to use the latest available information about each verifier, combining the information from all sources.
+
+When merging, information about a verifier found anywhere in the current _verifier lookup_ must be applied for all uses of a verifier (e.g. for revocation status).
+
+Wherever it is applicable and feasible, VOA technology backend implementations should implement "merging" to obtain a unified view of the latest information for each verifier.
+
+For example, if a verifier has been revoked, but only one of two copies of that verifier in VOA reflects this revocation, "merging" the two leads to stable visibility of the revocation status.
+Summarized, in technologies that implement merging, [load path] priority is not relevant (e.g. a non-revoked version in `/etc/voa/` of the same verifier never overrides the revoked version in `/usr/share/voa/`).
+The "latest normalized view" of a verifier is not impacted by load path priority.
+
+### Future compatibility
+
+No changes to the overall VOA structure are anticipated, which is why versioning is not encoded in the directory structure.
+However, if it turns out that a breaking change to the VOA specification is necessary a separate directory structure (e.g. `/usr/share/voa2/`) can be specified.
+
+Independently, breaking changes of the VOA format _within_ individual technologies may occur.
+In this case, versioned variants of the technology in question are added (e.g. `openpgp:2`).
+The `:` delimiter is chosen to signify VOA-specific versioning.
+This differentiates clearly between versions of the technology itself, as specified upstream (e.g. [OpenPGPv4], [OpenPGPv6]) and how these technologies are used in the context of VOA.
+
+## Identifiers
+
+_Signature verifiers_ are located in directory structures described by the [file hierarchy].
+Each of the following identifiers represents a subdirectory layer in that hierarchy.
+
+### OS
+
+The _os_ identifier is used to uniquely identify an Operating System (OS).
+
+The identifier relies on data provided by the ubiquitous [os-release].
+The following keywords are understood and their value format follows that established by [os-release] (i.e. no spaces and no characters outside of `0–9`, `a–z`, `"."`, `"_"` and `"-"` are allowed):
+
+- `ID`: name of OS (e.g. `arch` or `debian`)
+- `VERSION_ID`: the version of the OS (e.g. `1.0.0` or `24.12`)
+- `VARIANT_ID`: the variant of the OS (e.g. `server` or `workstation`)
+- `IMAGE_ID`: the image of an OS (e.g. `cashier-system`)
+- `IMAGE_VERSION`: version of the image (e.g. `1.0.0` or `24.12`)
+
+The values for these _parts_ must be provided, in the above order, as a colon-separated string that defines a specific _os_ identifier (e.g. `debian:12:server:company-x:25.01`).
+At least the `ID` _part_ must be set, an empty _os_ identifier is invalid.
+Trailing colons must be omitted for all _parts_ that are unset (e.g. `arch` instead of `arch::::`).
+
+The _os_ identifier may be extended with further _parts_ in an updated specification.
+Users encountering an _os_ identifier with more than 5 _parts_ should ignore these directories.
+
+Users of VOA (e.g. a Linux distribution creator) are free to store and retrieve verifiers in more or less specific _os_ identifiers, at their own discretion. VOA supports both generic identifiers (e.g. `arch`), and very specific identifiers (e.g. `fedora:41:workstation:cashier-system:1.0.0`).
+
+VOA is not flexible regarding _os_ strings:
+When an application looks up verifiers for e.g. `fedora:41:workstation`, then no verifiers from alternate _os_ strings are implicitly considered (e.g. from the more general location `fedora:41`).
+
+However, VOA libraries may offer an API that allows applications to use a list of _os_ strings (e.g. a list of both `fedora:41:workstation` and `fedora:41`).
+Such a call will use the combination of all verifiers found in both directory hierarchies.
+Applications that use VOA (via a VOA library) can therefore opt to pass a list of _os_ strings, if verifiers are known to be spread over different _os_ directories.
+VOA users can omit specific parts, e.g. `IMAGE_ID` and `IMAGE_VERSION`, if they know that these parts are not used in the [VOA hierarchy] for their _os_.
+
+### Purpose
+
+A _purpose_ combines a _role_ and a usage _mode_:
+
+- A _role_ acts as a trust domain (e.g. the "package" _role_ is used for signatures for package verification).
+- There are two _modes_ in which _signature verifiers_ can be stored in VOA:
+  - Verifiers used for direct artifact verification
+  - Verifiers that serve as _trust anchors_
+
+A _role_ and a _mode_ in combination form a _purpose_.
+
+_Trust anchor_ verifiers are always used to ascertain the validity of the associated _artifact verifiers_ (e.g. _trust anchor_ verifiers for the "package" _role_ are used to validate the _artifact verifiers_ for the "package" _role_).
+
+To use _signature verifiers_ in more than one _mode_ [symlinking] may be used.
+
+#### Directory naming
+
+_Purpose_ directories for direct _artifact verifiers_ are named `$role` (e.g. `package`), while _trust anchors_ are stored below directories named `trust-anchor-$role` (e.g. `trust-anchor-package`).
+
+Note that _trust anchor_ directory names always start with a "trust-anchor-" fragment.
+
+Either two or one _purpose_ directories may exist per _role_:
+
+- One directory which contains _artifact verifiers_ (e.g. `package`), and a second directory which contains the corresponding _trust anchors_ (e.g. `trust-anchor-package`).
+- Just one directory which contains _artifact verifiers_ (e.g. `package`).
+
+The _purpose_ directory name must not contain characters outside of `0–9`, `a–z`, `"."`, `"_"` and `"-"`.
+VOA implementations must not consider invalid directories and should raise a warning if such a directory is encountered.
+
+##### Two purpose directories: Verification relying on trust anchors
+
+For example, verifiers for the "package" _role_ may be stored in two _purpose_ directories:
+
+- `package` ("_artifact verifiers_ of actors who are designated for package signing"), and
+- `trust-anchor-package` ("_trust anchors_ for _artifact verifiers_ used for package signing")
+
+In this scenario, verification based on _trust anchors_ is performed.
+
+##### One purpose directory: Direct verification
+
+In another example, verifiers for the "image" _role_ may be stored in only one _purpose_ directory:
+
+- `image` ("_artifact verifiers_ of actors who are designated for OS image signing")
+
+In this scenario, only direct verification can be performed.
+
+#### Roles as trust domains
+
+Having distinct _roles_ allows the use of separate sets of _signature verifiers_ per role.
+This is useful if different actors are expected to issue signatures for each _role_.
+Thus, each _role_ acts as a trust domain, e.g. the "package" _role_ is used for signatures for packages, while the "image" _role_ is used for signatures for OS images.
+The work on each of the _roles_ may be performed by different teams, using different verifiers.
+
+#### Standard roles
+
+The standard roles defined by the VOA specification are:
+
+* **package**: Verifying signatures for packages
+* **repository-metadata**: Verifying signatures for repository metadata
+* **image**: Verifying signatures for OS images
+
+The corresponding _purpose_ directories are:
+
+- `package` and `trust-anchor-package`
+- `repository-metadata` and `trust-anchor-repository-metadata`
+- `image` and `trust-anchor-image`
+
+The above list of standard roles can be extended by users of VOA.
+
+For more in-depth explanation on the use of the _purpose_ identifier, see the [examples] section.
+
+### Context
+
+The **context** identifier allows defining specific verifiers for a particular context within an [os]'s [purpose].
+
+The _context_ identifier allows modelling finer grained trust domains within a _purpose_.
+This can be necessary if different actors are responsible for signing in subsets of a _purpose_.
+
+Specific examples for **context** are
+
+- the name of a specific software repository when certificates are used within the **package** [purpose] (e.g. `core` for a repository named "core")
+- how an OS image is used within the **image** [purpose] (e.g. `installation-medium`, `virtual-machine`)
+
+If no specific context is required, the **context** directory `default` must be used.
+
+Analogous to the os-release data, **context** strings may only contain `[a-z]`, `[0-9]`, `_`, `.` and `-`.
+
+For more in-depth discussion on the use of the _context_ identifier, see the [examples] section.
+
+### Technology
+
+The following sections outline specifics about the supported technologies.
+
+In VOA, technology-specific backends implement the logic for each supported technology.
+Different parties can implement different backends and this document outlines the approach, including the commonalities between all technology backends.
+
+The details of verifier [load logic] is defined per technology to leverage the individual features and strengths of each technology, while offering semantics that are closely shared between all VOA technologies.
+
+Currently, only the OpenPGP technology is specified in detail.
+Further specification work is required before implementing further technologies.
+
+Each technology specifies file suffixes for verifiers.
+Note that these suffixes may overlap.
+
+A _technology_ directory name must not contain characters outside of `0–9`, `a–z`, `"."`, `"_"` and `"-"`.
+VOA implementations must not consider invalid directories and should raise a warning if such a directory is encountered.
+
+#### OpenPGP
+
+[OpenPGP] is a widely adopted decentralized system for signature verification and user authentication.
+
+This technology is named `openpgp` in the VOA structure.
+
+Most Linux distributions use OpenPGP, often combined with _PGPKI_ (aka. [Web of Trust (WoT)]), in which chains of trust are evaluated during signature validation.
+
+OpenPGP has a rich model of validity, both for certificates and signatures.
+Certificates can expire or be revoked by the key holder (via [OpenPGP certificate revocation]).
+Similarly, [third-party identity certifications] (as used in the _PGPKI_) can be revoked by their issuers (using [OpenPGP signature revocation]).
+
+OpenPGP VOA backends are expected to handle validity.
+In particular they should take expiration and revocation into account in all layers of the [VOA hierarchy], by implementing [merging] semantics.
+For example, even if a revoked version of a certificate is overlaid by an unrevoked version, the revocation must be considered.
+
+OpenPGP VOA backends implementations must offer appropriate facilities to calculate the validity of _artifact verifiers_.
+This may include:
+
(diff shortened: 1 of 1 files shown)


specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,561 @@
...
+
+However, in other technologies, more complex [merging] logic may apply, that effectively combines information from different representations of one verifier.
+
+Additionally, the concept of [masking] applies for all technologies.
+
+#### Overriding
+
+In technologies that use overriding, verifier files shadow other variants of the same verifier based on filename equality.
+
+For example, `/etc/voa/fedora/packages/default/ssh/buildserver1.pub` overrides `/usr/share/voa/fedora/package/default/ssh/buildserver1.pub`.
```
xsuchy, line 158: You do not define which directories have what priority. Therefore, you can end up with various implementations.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,561 @@
...
+### Load paths
+
+VOA defines a list of _load paths_ for system mode and user mode, from which a hierarchy of _signature verifiers_ can be retrieved.
+
+Load paths when in system mode:
+
+- `/etc/voa/`
+- `/run/voa/`
+- `/usr/local/share/voa/`
+- `/usr/share/voa/`
```
xsuchy, line 113: Will all files be in one directory? Take a look at `distribution-gpg-keys` package, where there are thousands of keys. If all of them are in one directory, it would be a pain.

dvzrv: No, the signature verifiers are not all in one directory.

See e.g. https://github.com/uapi-group/specifications/pull/134/files#diff-c79d1da1ef2fcfffc28ac7308505535eac0942d086d54a990553374ac81fed00R98 where the hierarchy is first mentioned and from where the various subdirectory layers can be accessed.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Do you think it should be made more clear than what is mentioned in https://github.com/uapi-group/specifications/pull/134/files#diff-c79d1da1ef2fcfffc28ac7308505535eac0942d086d54a990553374ac81fed00R148?

Granted, the priority is not mentioned explicitly in load paths or load logic, albeit that would probably be good indeed.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,561 @@
...
+
+Load paths when in user mode:
+
+- `$XDG_RUNTIME_DIR/voa/`
```
poettering, line 117: to be fully symmteric `$XDG_CONFIG_DIRS/voa/` would have to come first I guess?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,561 @@
...
+- `$XDG_DATA_HOME/voa/`
```
poettering, line 118: should probably be `$XDG_DATA_DRS/voa/` no?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
xsuchy: nod. E.g. what has higher priority: `/usr/share/voa/` or `/etc/voa/`?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Hmm, good point, but would we then not want to have XDG_DATA_DIRS in addition to XDG_DATA_HOME?

Looking at [the definition](https://specifications.freedesktop.org/basedir-spec/latest/#variables) I'm under the impression, that XDG_DATA_HOME has higher ordering than what is provided via XDG_DATA_DIRS?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
poettering: migth make sense to reference https://uapi-group.org/specifications/specs/configuration_files_specification/

that spec suggests /etc overrides both /run and /usr, and /run overriding /usr

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
Foxboron: @poettering `XDG_DATA_DIRS` is an array of data directories similar to `$PATH`. So `XDG_DATA_HOME` is correct? You could include that `XDG_DATA_DIRS` should also be looked at.

Foxboron: > If `$XDG_DATA_DIRS` is either not set or empty, a value equal to `/usr/local/share/:/usr/share/` should be used.

I think the specc needs to be clarified a bit more if you want to include `XDG_DATA_DIRS`.

Reff; https://specifications.freedesktop.org/basedir-spec/latest/

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Ah, right, that ordering should be adjusted! :thinking:

As an aside: We were wondering about whether voa should live inside of XDG_CONFIG_DIRS and decided against it as it basically contains data.
I guess there are good reasons for either XDG_CONFIG_DIR and XDG_DATA_HOME?

FWIW, XDG_CONFIG_DIRS should fall back to `/etc/xdg` (which we are not defining/using system-wide).
Meanwhile XDG_DATA_DIRS should fall back to  `/usr/local/share/:/usr/share/` (which we are using system-wide).

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
poettering: see section 4 of the basedir spec. specs should reference $XDG_DATA_DIRS/foo/bar and that shall implicitly mean $XDG_DATA_HOME too, and if neither variable is set the default. i'd just use that. or in other words: no need to specify XDG_DATA_HOME among the search path list explicitly then.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Good point. Makes sense to weave that in!

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
poettering: (i mean, you can make this explicit in your spec if you want, but it's kinda redundant to mention `$XDG_DATA_HOME` explicitly if you use the `$XDG_DATA_DIRS/foo/bar` wording, that's all I am saying. You could also add `as per section 4 of xdg spec` or so in your spec

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
poettering: i think it might be totally ok if people manually configure some addition key if they like, i.e. let's say you have a user level package manager of sorts, and want to register a copr-style set of keys for it, it might be appropriate to do that as part of the config.

the way i'd think about XDG_CONFIG_DIRS vs. XDG_DATA_DIRS is that the former is more user territory, where they manually make changes and the latter is more app/managed territory where programs more autmatically make changes. and since both are valid I'd allow both.

I think the /etc/xdg thing is a major brokeness in the basedir spec btw. in systemd we usually say we support it, but never docuemnt/announce that we do, and just use the /etc/ wherever the spec says /etc/xdg as primary place for stuff.

poettering: one thing to think about: when i have some code in user mode, and looking for key meterial it shall sue for authenticating some resources, should it actually *just* look in the per-user dirs? or should it look in both the per-user and per-system dirs? if the latter is the case things are simple, because xdg basedir (with the exception of the /etc/xdg/ weirdness) pushes you towards it.

however, if the two scopes should not be overlapping, then you'd need a different top-level entrypoint dir name for the voa stuff, for example "voa-user" instead of "voa" or so, so that you look into /usr/share/voa-user instead of /usr/share/voa if running in user scope mode.

but frankly I would not bother, and simply declare that the per-user keyrings should generally be a superset of the per-system keyrings, and if people disagree with that they could use /dev/null masking in their per-user dirs to block out keys provided in the per-system dirs.

poettering: (just to mention this, the whole reason why units in systemd are dropped in /etc/systemd/system and /etc/systemd/user is precisely this: to separate the scopes entirely. I think separating unit search dirs fully is essential, but here for the key material i think it's unnecessary: after all we have a hierarchy of trust: there's no point in trusting some user level stuff if you cannot trust the system your user-level stuff runs on. hence, you always implicitly trust the system when you trust your user code, and hence there's no point in separating the search scopes on disk. hence i am pretty sure that user-scope should just mean we *add* a few more dirs to the search path order compared to system-scope, but we do not remove any)

dvzrv: > should it actually just look in the per-user dirs? or should it look in both the per-user and per-system dirs?

We have been thinking about that as well and did not come to a clear conclusion.
That's why we erred on the side of "only make per-user verifiers available in user-mode" (but without introducing a separate directory structure only for user-mode in a system-wide location).
I agree though, that it would be more convenient to re-use the system-wide verifiers and e.g. only mask those that are unwanted (instead of copying those  that are wanted into place).

I am not against changing this by relying on e.g. XDG_DATA_DIRS which by default would use the correct system-mode locations.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Added in https://github.com/uapi-group/specifications/pull/134/commits/20dcf7432b17d7bca04905eb3ddc8c64122e84fb

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Added all the XDG dirs in https://github.com/uapi-group/specifications/pull/134/commits/20dcf7432b17d7bca04905eb3ddc8c64122e84fb

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: I've added support for XDG_CONFIG_DIRS and XDG_DATA_DIRS in https://github.com/uapi-group/specifications/pull/134/commits/20dcf7432b17d7bca04905eb3ddc8c64122e84fb

DaanDeMeyer:
Oh shit I was scrolling on the github app on my phone and fat fingered the close button

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+The _File Hierarchy for the Verification of OS Artifacts (VOA)_ defines a generic approach for storage and retrieval of _signature verifiers_, supporting a wide range of cryptographic technologies.
+
+VOA aims to cover all verification needs around OS artifacts.
+
+Unlike e.g. Mozilla's Network Security Services ([NSS]), which concerns itself with secure network communication, including a curated set of X.509 CA certificates (e.g. often provided on Linux distributions via `/etc/pki` or `/etc/ca-certificates`), VOA focusses on the verification of artifacts with support for a diverse set of cryptographic technologies.
+
+## Terminology
+
+This text uses the generic term _"signature verifier"_ for cryptographic objects that are used to verify signatures.
+In practice, those may be for example X.509 certificates, or OpenPGP public keys.
```
dkg, line 29:
```suggestion
In practice, those may be for example X.509 or OpenPGP certificates.
```

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+(e.g. SSH, minisign, signify)
+
+Keys are "atomic" and don't have formal relationships with each other: A key issues a number of signatures, which can be verified.
+
+These systems use "set-style trust": Users choose to rely on some _set_ of keys/verifiers, and are willing to accept signatures made with those.
+
+This model doesn't require or support use of _trust anchors_ in VOA.
+In this model, the set of _artifact verifiers_ is considered implicitly trusted.
+
+#### Decentralized trust
```
dkg, line 58: I would avoid the term "trust" here (and in the next section).  No one really knows what it means, and using it here is more likely to cause friction and confusion than illumination.

For example, OpenPGP can be used "hierarchically", and X.509 "trust anchors" themselves often are cross-signed, forming a mesh.

The real questions this section and the subsequent one have are about _delegation_, _discovery_, and _directness_.

In the questions below, i use the more colloquial "X is allowed to sign Y" to mean "an object Y would be considered acceptable if it carries a verifiable signature from X".

- `delegation`: If a cryptographic object is allowed to sign an OS artifact, is it also allowed to delegate it's authority to a distinct cryptographic object?
- `discovery`: If any delegation is possible, is it acceptable to discover novel cryptographic objects during the course of OS artifact verification?  Or, does the set of acceptable signers need to be known in advance?  in the latter case, the "delegation" mechanisms serve as an ability to effectively revoke a known signer.
- `directness`: Are there any relevant cryptographic objects that *cannot* sign an OS artifact, but *can* delegate the authority to do so?  for example, an X.509 "trust anchor" CA certificate that does not have a codesigning eKU cannot *directly* sign an OS artifact, it would need to mint a new direct signer.

There is an additional fourth prong that might be interesting for some use cases, which is "corroborability" -- such as a scenario where you'd like to have multiple signers before a given artifact is approved.  In practice, i don't see anyone doing that right now (but see also https://gitlab.com/dkg/openpgp-stateless-cli/-/issues/104 for additional discussion)

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+- `$XDG_CONFIG_HOME/voa/`
+- the `./voa/` directory in each directory defined in `$XDG_CONFIG_DIRS`
+- `$XDG_RUNTIME_DIR/voa/`
+- `$XDG_DATA_HOME/voa/`
+- the `./voa/` directory in each directory defined in `$XDG_DATA_DIRS`
+
+### Symlinking
+
+Symlinks can be used in the VOA hierarchy to point to files or directories below one of the [load paths].
+Symlinking to files or directories external to the load paths is not supported.
```
dkg, line 136: why are symlinks rejected?  should an evaluator that discovers a symlink provide a warning explaining the rejection?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+
+### Masking
+
+Individual _artifact verifiers_ and _trust anchors_ may be masked using a symlink to `/dev/null`, independent of [technology].
```
dkg, line 140: Why are only individual artifact verifiers and trust anchors maskable?  is there a reason that i wouldn't be able to mask an entire OS/purpose combination?  or does that not make sense -- a given verifier already knows the full os,purpose,context,technology tuple and so masking at these grouping levels doesn't do anything reasonable.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+### Avoiding duplication of data in the directory structure
+
+Symlinks are explicitly supported when forming VOA structures.
+In many cases, it will be desirable to make heavy use of symlinks, to form the appropriate directory structures while avoiding duplication of data.
+However, the [VOA hierarchy] is intended as a self-contained data structure that specifies how to authenticate signatures.
+Therefore, symlinks in VOA should only point to other files or directories in any of the [VOA hierarchy] paths.
+Symlinks may be relative or absolute.
+Using files external to the VOA structure is not permitted and should lead to an error.
+
+The use of hardlinks in the VOA structure is not supported and should lead to an error.
```
dkg, line 507: how is a hardlink supposed to lead to an error?  does the VOA implementation read `st_nlink` and yield an error if it is > 1?  what do we gain from this constraint?

dkg:
Thanks for this!  I appreciate the thought that has gone into it, and i think it would be an improvement over the current scattershot status quo.  I worry that it is so intent at being as general as possible that there are too many knobs in it (e.g. the four-layer hierarchy is a bit intimidating) but i'm also not sure what i would cut, so maybe it's at the appropriate level of complexity.

dkg:
Thinking about this a bit more, i have an additional concern about the "merging" section.

I think we all know what it means when there are two copies of an OpenPGP certificate that share the same primary key.  But what should happen if two identically named _signature verifier_ OpenPGP certificates in two parts of the search tree each contains a certificate that has a *different* primary key from the others?  What does "merging" look like in that case? is it an error?   Should both certificates found be honored?

[`sop merge-certs`](https://www.ietf.org/archive/id/draft-dkg-openpgp-stateless-cli-13.html#name-merge-certs-merge-openpgp-c) is an example of a standardized OpenPGP certificate merging mechanism that has multiple existing implementations.  But note that `sop merge-certs` is (deliberately) asymmetric:  the `CERTS` objects on standard input are retained and augmented with anything found from the other `CERTS` objects identified on the command line _as long as they share a primary key_ (or are [otherwise equivalent](https://datatracker.ietf.org/doc/draft-ietf-openpgp-replacementkey/));  But any `CERTS` object identified on the command line which is not equivalent with the `CERTS` on stdin is discarded.  This was a deliberate choice, so that an application using `sop` to manage an OpenPGP certificate could safely apply updates (revocations, new subkeys, UIDs, expirations, etc) found from dubious origins (e.g., keyservers) and *not* worry about accidentally incorporating an entirely new OpenPGP certificate should the dubious origin decide to feed such a thing to the client.

I think the "merging" that's asked for in this draft spec is intended to be more symmetric than `sop merge-certs`.  That's fine, but it's worth documenting what's expected here.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
Torxed: > a scenario where you'd like to have multiple signers before a given artifact is approved.

It is used in real world scenarios, but they are usually limited to high assurance environments. It's perhaps a niche market, but having the option to do so is not a bad idea from a design standpoint - as it would allow for future use without changes.

> I would avoid the term "trust" here

I personally believe the word "trust" is often used in these contexts, so I don't see the harm in using it here as it's part of the industry standard lexicon and is described in the text body. But the word does have different meanings depending on which technology context it's being used in.

But describing the mechanics as you did might be a good idea, as they are key parts of the trust process.

(Hope I understood the remark correctly, otherwise I'm out fishing here - if so I apologize)

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+
+### Masking
+
+Individual _artifact verifiers_ and _trust anchors_ may be masked using a symlink to `/dev/null`, independent of [technology].
+This constitutes the only allowed exception to the rule of not symlinking to files or directories external to the load paths.
+
+These symlinks are only expected to be used in writable and persistent load paths (that is, only below `/etc/voa/` when in system mode, or below`$XDG_DATA_HOME/voa/` when in user mode).
+Masked verifiers are not considered when retrieving from any of the VOA [load paths].
+
+Masking _artifact verifiers_ and _trust anchors_ may be desirable in situations where a verifier is considered compromised, but a revocation for it can not be provided to users in time.
```
Torxed, line 146: I wonder if we should clarify behavior in read-only file systems more, would these constraints be problematic? In conjunction with https://github.com/uapi-group/specifications/pull/134/files#diff-c79d1da1ef2fcfffc28ac7308505535eac0942d086d54a990553374ac81fed00R115 there would be ways to work with such filesystems. But symlinking to `/dev/null` might be problematic.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+
+#### Overriding
+
+In technologies that use overriding, verifier files shadow other variants of the same verifier based on filename equality.
+
+For example, `/etc/voa/fedora/packages/default/ssh/buildserver1.pub` overrides `/usr/share/voa/fedora/package/default/ssh/buildserver1.pub`.
+
+Overriding depends on file name equality and directory structure.
+
+For example, `/etc/voa/fedora/packages/default/ssh/buildserver2.pub` does not override `/usr/share/voa/fedora/package/default/ssh/buildserver1.pub`, but instead adds an additional _signature verifier_.
```
Torxed, line 168: These are human readable nomenclatures, but what if the fingerprint of the certificate/key is identical? Should the fingerprint be taken into consideration instead? This would also avoid accidental duplicates.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+Great care has to be taken to ensure the naming scheme has the desired overriding behavior (e.g. to make sure that a revoked version of an SSH public key overrides a non-revoked version of the same SSH public key found in a different load path).
+
+#### Merging
+
+Some technologies use a business logic that "merges" different versions of the same _signature verifier_ (instead of applying overriding semantics).
+The purpose is to use the latest available information about each verifier, combining the information from all sources.
+
+When merging, information about a verifier found anywhere in the current _verifier lookup_ must be applied for all uses of a verifier (e.g. for revocation status).
+
+Wherever it is applicable and feasible, VOA technology backend implementations should implement "merging" to obtain a unified view of the latest information for each verifier.
```
Torxed, line 179: What would be the outcome if a OpenPGP key has expired in one signature verifier entry, and another version has been modified with a new expiry date. I am guessing the new expiry date takes prioerity? Or would it be counted as revoked/expired still?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,568 @@
...
+
+For example, if a verifier has been revoked, but only one of two copies of that verifier in VOA reflects this revocation, "merging" the two leads to stable visibility of the revocation status.
+Summarized, in technologies that implement merging, [load path] priority is not relevant (e.g. a non-revoked version in `/etc/voa/` of the same verifier never overrides the revoked version in `/usr/share/voa/`).
+The "latest normalized view" of a verifier is not impacted by load path priority.
+
+### Future compatibility
+
+No changes to the overall VOA structure are anticipated, which is why versioning is not encoded in the directory structure.
+However, if it turns out that a breaking change to the VOA specification is necessary, either a separate directory structure (e.g. `/usr/share/voa2` etc.), or a new subdirectory such as `/usr/share/voa/v2` can be specified.
```
Torxed, line 188: Adding `v2` to the VOA path, would break the assumption that `$os` should follow the root, for instance implementations would not be able to assume `os` lives under the root, and would cause multiple implementations/checks "if v2 exists" in: `/usr/share/voa/$os/$purpose/$context/$technology/`.

Perhaps it would be wise to have the version number in the path from the get-go? If we anticipate/recommend doing it for v2 already in this description :)

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Thanks for raising this!

Masking directory structures would indeed be very complex to reason about and is explicitly not supported.
We will adapt the text accordingly!

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: What this sentence was meant to convey is that hardlinks are confusing and should not be used.

Not checking for this explicitly in a VOA implementation is fine though and this should not be a hard requirement for implementers.

We will adjust the text accordingly.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Ah yes, similar to how `systemctl mask` can work on `--system` or `--runtime`, we can both use e.g. `/etc/voa/` or `/run/voa/`.

We have clarified the masking section and added a section for implementers about this.

dvzrv:
> I think the "merging" that's asked for in this draft spec is intended to be more symmetric than `sop merge-certs`. That's fine, but it's worth documenting what's expected here.

Thanks! That is a good point!
When revising the OpenPGP and SSH technology section we added additional constraints:
Implementations must not use _signature verifiers_ where the fingerprint of the file name does not match the fingerprint of the contained certificate/public key (and instead emit a warning).

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Thanks for the great input!

We have adjusted the classification section to make it more clear that the various technologies are described in general terms.
In VOA they are used in a very constrained subset of their specific capabilities.

In the file hierarchy section we addressed the retrieval of _signature verifiers_ being out of scope for this document (but added a note for implementers about this topic at the end).

The OpenPGP technology section has been extended to clarify the default _trust amount_ and _trust depth_ values.

Closing, we have also added an implementer's section on "threshold signing" (aka. "corroborability").

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,615 @@
...
+- Rely on _trust anchors_ which delegate to one or more _artifact verifiers_.
+  Anchors must be explicitly defined in the relevant "trust-anchor" VOA directory (see [purpose]).
+
+#### Hierarchical delegation
+
+(e.g. X.509, SSH+CA)
+
+Other systems have an inherent assumption of hierarchical delegation, formalized by cryptographic certifications.
+In such systems, globally or locally accepted _trust anchors_ are used as starting points in signature validation.
+Certificates and the delegation relations between them form a tree structure.
```
baloo, line 78: This is a bit of a nitpick, but this is not a tree, but more an acyclic graph (and even then, maybe the acyclic is wrong?)

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,615 @@
...
+
+```
+/usr/share/voa/fedora:41/image/installation-medium/openpgp/62cdb7020ff920e5aa642c3d4066950dd1f01f4d.openpgp
+```
+
+The VOA library will also check for verifier files in the other layers, but in in this example, there are none.
+
+The VOA library has found one _artifact verifier_ file (`/usr/share/voa/fedora:41/image/installation-medium/openpgp/62cdb7020ff920e5aa642c3d4066950dd1f01f4d.openpgp`) and proceeds to use it to verify the installation medium file.
+
+## Considerations for implementers and users
```
baloo, line 525: One thing I'd like to see is a recommendation about signing-date.

At least with x509, the date considered when checking an x509 path should be the time when the signature was emitted. (Ideally that date comes from an TSA (Timestamp authority)).

Otherwise a signature may become invalid once a single cert in the x509 path expires.

I believe this should probably also apply to pgp as keys may expire, but that also should apply to ssh+ca.

dvzrv: Good point. We'll add something about that topic!

Related, would you be interested in revising the [X.509 technolgy section](https://github.com/uapi-group/specifications/pull/134/files#diff-c79d1da1ef2fcfffc28ac7308505535eac0942d086d54a990553374ac81fed00R411-R429) and providing improvements?
It would help us a great deal in getting that section out of draft mode and functional.
In my mind having at least two fully specified technologies would make this PR merge-worthy.

baloo: About that:
x509 only carries information about public keys and trust delegated to them (and their limitation). This is the focus of that document, no comment here.

But x509 does not sign anything in itself. It doesn't standardize the signature scheme.

Is this left to the implementation? Or should go more specific?
I guess we could either go with a JWS, or most likely towards PKCS7/CMS? The latter is the one used for signing EFI payloads.
And CMS is very flexible! Which is great, since this we can add more than one signature to an object, carry a signature date, can provide various alternative x509 paths, provide counter signatures (Signed timestamps, ...) and other nuggets.
That also makes the format somewhat hard to work with/comprehend.

I would really advise against going with straight ecdsa or RSA signature (which do not carry a signature date).

But again, signature format doesn't seem to be the purpose of this spec, and I'm not sure how much of it fits in.

Torxed: I do like CMS too, but I'm a bit biased :)

baloo (approved):
Thanks!

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: > One thing I'd like to see is a recommendation about signing-date.

We have added a subsection that suggests using TSAs if the technology integrates with it.

> But x509 does not sign anything in itself. It doesn't standardize the signature scheme.
>
> Is this left to the implementation? Or should go more specific?
> I guess we could either go with a JWS, or most likely towards PKCS7/CMS? The latter is the one used for signing EFI payloads.
> And CMS is very flexible! Which is great, since this we can add more than one signature to an object, carry a signature date, can provide various alternative x509 paths, provide counter signatures (Signed timestamps, ...) and other nuggets.
> That also makes the format somewhat hard to work with/comprehend.

Being more specific here would be good (e.g. what types of signatures are supported, how do the certificates need to be structured, what kind of certificates are allowed).
It would be much appreciated to get more specific improvement suggestions for the X.509 section. :pray: 

I am no expert on PKCS#7/CMS, but to my knowledge @poettering would be happy to see PKCS#7 support for the various EFI and image signing integrations and I hope to be able to grab his attention for it once more :sweat_smile: 

> I would really advise against going with straight ecdsa or RSA signature (which do not carry a signature date).
>
> But again, signature format doesn't seem to be the purpose of this spec, and I'm not sure how much of it fits in.

I agree. Thanks for pointing this out!
We will at least add this as a minimum requirement for the X.509 technology section.
At least providing "rough" guidelines around the signature format seems definitely in scope for the X.509 technology section if the format allows for a lot of flexibility.
(FWIW, we ran into similar issues in the OpenPGP technology section, as the technology can be used quite flexibly.)

hko-s: This document tries to strike a balance of allowing the use of different approaches/technologies, while also defining enough specifics of a common approach to hopefully be helpful.

Good question how this interacts with the signature format (which this document indeed mostly sidesteps).

Depending on the technology, it's kind of clear what signature format is expected (e.g. OpenPGP signatures are part of the format). But where it's not clear, I think VOA should gesture towards reasonable approaches, but ultimately let library and application developers do what they must.

In the case of X.509 I don't have any personal experience that would help me judge how and where to gesture (and neither does @dvzrv, which is why both of us are very happy to have input about the domain).

That said, gesturing towards PKCS7/CMS sounds plausible to me, if multiple people with experience in the domain agree that it is.

dvzrv: I've added a recommendation for using signature formats, that include signature creation timestamps to the X.509 section.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: We have added a clarification.

Do note, that for the purpose of this document we are sometimes referring to subsets of functionality of certain technologies (to make a point about how this technology is used in the context of VOA).

In case you would still like to see the text improved, don't hesitate to reopen!

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,639 @@
...
+In VOA, by convention, OpenPGP certificates must be stored in individual files and provided in ASCII armored form.
+
+These files must be named by certificate fingerprint (in lowercase hex notation) with the file ending `.openpgp` (e.g. `d8afdda07a5b6edfa7d8ccdad6d055f927843f1c.openpgp`).
+
+OpenPGP VOA backends must reject files where the certificate fingerprint of the file name does not match the actual fingerprint of the contained OpenPGP certificate's primary key and should emit a warning if such a file is encountered.
+
+By default, in VOA, OpenPGP certificates that act as _trust anchors_ are considered with a _trust amount_ of 40 at a _trust depth_ of 1.
+More complex delegation setups are possible, but must be implemented using an application specific configuration mechanism.
+
+#### SSH
```
wiktor-k, line 383: FYI, I've mentioned using `ssh-keygen -Y verify` in [this e-mail](https://lists.mindrot.org/pipermail/openssh-unix-dev/2025-April/041900.html). (Currently the lack of support for multiple input files is a bit limiting).

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,639 @@
...
+Currently, OpenPGP and X.509 are widely adopted.
+
+As of this writing, no technology-agnostic, standardized location for the distribution of cryptograpic material that serves as verifier for digital signatures exists.
+This leaves consumers to either do guesswork, or rely on proprietary, stateful or technology-specific keystore formats and per-application locations.
+
+The _File Hierarchy for the Verification of OS Artifacts (VOA)_ defines a generic approach for storage and retrieval of _signature verifiers_, supporting a wide range of cryptographic technologies.
+
+VOA aims to cover all verification needs around OS artifacts.
+
+Unlike e.g. Mozilla's Network Security Services ([NSS]), which concerns itself with secure network communication, including a curated set of X.509 CA certificates (e.g. often provided on Linux distributions via `/etc/pki` or `/etc/ca-certificates`), VOA focusses on the verification of artifacts with support for a diverse set of cryptographic technologies.
```
poettering, line 24: not that it matters, but in systemd docs we generally insist on suffixing paths that refer to dirs with a `/`, simply to visually communicate that we are talking of a dir here.

(just an idea, no need to follow systemd's rules here, but we have them for a reason, a reason I think makes sense)

poettering:
BTW, I think it might be interesting to make sure that VOA is good enough to be a good source to generate UEFI SecureBoot "db" files from. so that one can ship uefi x509 signing keys in VOA, so that some future tool could then optionally pick up those certs from the VOA tree and convert them into an unified UEFI ESL which could be written to a "db" uefi variable. 

After all, besides their native package signining for each distribution UEFI SecureBoot signing is the other kind of signing everyone needs to think about. It would make a ton of sense to me, if those keys could be stored the same way, so that we can process them the same way. In particular as in many scenarios one probably wants to combine keys from msft, with keys from the distros, with possibly keys from the local admin into a combined ESL in different ways, hence being able to read them from VOA and turning them locally into an ESL would make sense.

In my understanding UEFI ESL is just a concatenation of two kinds of objects: x509 certificates and of "raw" hashes of objects. I figure VOA is fine with the former, but some thought needs to go into how to encode the "raw" hashes.

The efi specs about all of this are here:

https://uefi.org/specs/UEFI/2.11/32_Secure_Boot_and_Driver_Signing.html#signature-database

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,639 @@
...
+- the `./voa/` directory in each directory defined in `$XDG_CONFIG_DIRS`
+- `$XDG_RUNTIME_DIR/voa/`
+- `$XDG_DATA_HOME/voa/`
+- the `./voa/` directory in each directory defined in `$XDG_DATA_DIRS`
+
+### Symlinking
+
+Load paths are constrained to self-contained locations on a host as they provide vital data for the integrity and verification of all components on a system.
+However, symlinks can be used in the VOA hierarchy to point to files or directories below one of the [load paths] in descending priority.
+Symlinks to files or directories below ephemeral load paths (i.e. `/run/voa/` and `$XDG_RUNTIME_DIR/voa/`) are prohibited, as they would lead to dangling references.
```
poettering, line 150: *possibly* lead to dangling references

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,639 @@
...
+
+As an example, symlinks to files below the same load path or to another load path with lower priority may be used to deduplicate the use of a single _signature verifier_ for multiple use-cases.
+Additionally, using symlinks allows to automatically keep _signature verifiers_ in sync with canonical upstream data.
+
+Symlinking to files or directories external to the load paths is prohibited.
+VOA implementations must not consider symlinks to files outside of the specified load paths and should raise a warning if such symlinks are encountered.
```
poettering, line 156: you probably want to say here that symlinks must be relative, may not be absolute.

(that's kinda handy so that people can bind mount a voa tree into a container and it works)

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,639 @@
...
+#### Roles as trust domains
+
+Having distinct _roles_ allows the use of separate sets of _signature verifiers_ per role.
+This is useful if different actors are expected to issue signatures for each _role_.
+Thus, each _role_ acts as a trust domain, e.g. the "package" _role_ is used for signatures for packages, while the "image" _role_ is used for signatures for OS images.
+The work on each of the _roles_ may be performed by different teams, using different verifiers.
+
+#### Default roles
+
+The default roles are:
```
poettering, line 302: do you mean "default"? 

don't you want to say: "Standard roles defined by this spec"?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
poettering: so, don overestimate my understanding of the x509 ecosystem. I only grok some parts of it, for the stuff I actually needed to work with. I have worked with pkcs7 in the past, and it seems fine to me and was an obvious choice. I have no idea whether other stuff might be better, but I'd always focus on the conservative choice here, which pcs7 seems to be?

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,639 @@
...
+Each [technology] may handle validity differently.
+Generally, signature validity should be considered at the time of signature creation.
+The _signature verifier_ must be valid at creation time.
+If the _signature verifier_ has expired after signature creation, this does not impact signature validity.
+Only in the case of revocations, that indicate a key material compromise, all signatures by the key in question should be considered invalid, regardless of creation time.
+
+#### Use of Time Stamp Authority
+
+Some technologies support the use of _Time Stamp Authorities_ (e.g. based on the [time stamp protocol]).
+It is recommended to use timestamping services to validate signature creation time, where possible.
```
poettering, line 588: I figure you want to emphasize *signature* here, because that's the key info here, no? i.e. not certificate timestamps, but signature timestamps.

poettering:
> No, they are the authenticode checksums of the EFI binaries. Not the raw hashes (they are useless).

well, the "raw" was referring to the fact that the hashes are not stored in any kind of relevant envelope, but are just raw hash values. I am fully aware that for the purpose of authenticating PE files the PE files are preprocessed, and certain parts removed from the data stream to hash. (I implemented authenticode hashing in systemd after all). 

I think from the perspective of VOA these hashes are really just raw hashes, i.e. come with very little metadata, except the algorithm/width. And I guess that's something we could include in the file suffix. Hence maybe we should augment the spec to allow encoding raw hashes with technology set to "raw-hash", and then use file suffix such as ".sha256" and ".sha512" and so on, and expect hex-encoded hash values in those files.

I think accepting simple direct hashes (instead of doing asymmetric crypto) makes a ton of sense in many usecases beyond Authenticode. For example, I am working with image based OSes where one base image is combined with a set of extension images, but all of these extension images are built at the same time as the base image, and while each deployment might only use a subset of them, they will be only combined from the same build version. For cases like that it makes no real sense to do asymmetric crypto, it's entirely sufficient to just have a VOA tree where simple raw hashes pin the extension images by their hash. hence, beyond the authenticode usecase I think it would make sense to have a generic concept for simple hashes in place, even if package managers are not going to use that.

> Efforts to distribute this in a more structured way is done in the `secureboot_object` project from Microsoft. https://github.com/microsoft/secureboot_objects/tree/main
> 
> See the JSON schema they use.
> 
> https://github.com/microsoft/secureboot_objects/blob/main/PreSignedObjects/DBX/dbx_info_msft_1_14_25.json
> 
> https://github.com/microsoft/secureboot_objects/blob/main/PreSignedObjects/DBX/HashesJsonSchema.json

interesting. This contains considerably more fields and info than the final ESL contains though, afaics, no?

question is whether to bother in the context of VOA with the additional fields or just focus on the data strctly needed for the ESL stuff...

i mean, fundamentally: while x509 is very extensible, the openssh key format as another example is not, afaik. these additional json fields that that msft project enhances the authenticode hash list with are certainly useful data, but i figure such additional metadata would not just be useful for the hashes of authenticode stuff, but also maybe for openssh or other formats. hence maybe one day we actually want some metadata "sidecar" for any kind of key we include in the VOA tree, similar in style to what msft adds to the authenticode hashes? But maybe leave that for the future?

(note though that the json stuff msft adds to the authenticode hashes is fundamentally different from the x509 metadata fields though: that data is actually protected by a signature of the x509 public key itself, hence you have trust from the x509 cert on itself. Which is great. The msft json stuff for the authenticode hashes is entirely unprotected though afaics, it's informational, and not trusted beyond maybe the git repository hash tree)

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
wiktor-k: This section should expand on the relation of the **context** and the SSH signature namespace.

Each SSH signature contains a namespace. [Documentation](https://github.com/openssh/openssh-portable/blob/e048230106fb3f5e7cc07abc311c6feb5f52fd05/PROTOCOL.sshsig#L53):

> The purpose of the namespace value is to specify a unambiguous interpretation domain for the signature, e.g. file signing. This prevents cross-protocol attacks caused by signatures intended for one intended domain being accepted in another. The namespace value MUST NOT be the empty string.

One way of connecting these two is specifying that SSH keys stored under given **context** directory are meant to verify signatures under namespace that is equivalent to the context. (Since both the context and the namespace are strings).

Note that the [`allowed_signers` format](https://man.archlinux.org/man/ssh-keygen.1.en#ALLOWED_SIGNERS), as used by OpenSSH uses `namespaces` keyword for restricting key usage ([example here](https://github.com/wiktor-k/ssh-signing/blob/main/allowed_signers.md?plain=1#L4)). Connecting these two would allow tools to convert VOA dirs into `allowed_signers` format.

(Note 2: git commit SSH signatures use `git` namespace)

A side-note: it'd be also good to specify the value of an SSH key comment. It could be an e-mail of the key-holder etc.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Thanks for finding the remaining outliers!
(all the other directory names already behaved as you described :sweat_smile: )

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: > One way of connecting these two is specifying that SSH keys stored under given context directory are meant to verify signatures under namespace that is equivalent to the context. (Since both the context and the namespace are strings).

We've added a note on this to the SSH section.

> A side-note: it'd be also good to specify the value of an SSH key comment. It could be an e-mail of the key-holder etc.

Do you mean that similar to the authorized_keys file format?

dvzrv:
As we believe that we have reached consensus on the current design of the VOA structure and the first technology (OpenPGP), we have now removed the remaining, yet to be fully specified technologies from this pull request.

The other technologies are now discussed in dedicated pull requests:
- SSH: https://github.com/uapi-group/specifications/pull/140
- X.509: https://github.com/uapi-group/specifications/pull/141
- minisign: https://github.com/uapi-group/specifications/pull/142
- signify: https://github.com/uapi-group/specifications/pull/143

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Let's continue in https://github.com/uapi-group/specifications/pull/140

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Let's continue the discussion on X.509 as VOA technology in https://github.com/uapi-group/specifications/pull/141

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
```diff
@@ -0,0 +1,548 @@
...
+- the name of a specific software repository when certificates are used within the **package** [purpose] (e.g. `core` for a repository named "core")
+- how an OS image is used within the **image** [purpose] (e.g. `installation-medium`, `virtual-machine`)
+
+If no specific context is required, the **context** directory `default` must be used.
+
+Analogous to the os-release data, **context** strings may only contain `[a-z]`, `[0-9]`, `_`, `.` and `-`.
+
+For more in-depth discussion on the use of the _context_ identifier, see the [examples] section.
+
+### Technology
```
dvzrv, line 340: This also needs a character set limitation (similar to OS).

dvzrv:
> I think a VOA implementation should refuse to load the verifier via the custom context path, in both cases.

I agree. I've adapted the text accordingly in edea396ecca460256e43ea68593b015e71114ced.
I'll have a look at the implementation in the coming days and we can have another look at this together on Wednesday.

specs/file_hierarchy_for_the_verification_of_os_artifacts.md
dvzrv: Thanks for clarifying this! We have adapted and simplified the symlinking section accordingly.

Foxboron:
Premature merge? Nobody approved this MR.

poettering:
Sorry, I gave my support via chat.

PR #135, opened 2025-02-13
bls: introduce "uki" and "uki-url" keys to the BLS spec

poettering:
This is used by https://github.com/systemd/systemd/pull/36314 to make networking boot work nicely within the BLS constructs.

The "uki" stanza also should have uses in a world where a single UKI shall be invoked by multiple locally defined menu entries, for example on nixos or ostree systems, where each commit shall synthesize a separate menu entry, often referencing the same UKI.

### DIFF
--- a/specs/boot_loader_specification.md
+++ b/specs/boot_loader_specification.md
@@ -279,6 +279,36 @@ The following keys are recognized:
 * `efi` refers to an arbitrary EFI program. If this key is set, and the system
   is not an EFI system, this entry should be hidden.
 
+* `uki` is a combination of the `linux` key, the `efi` key and Type #2 semantics: if this key is used the
+  specified path shall refer to a [unified kernel image (UKI)](unified_kernel_image.md). This informs the
+  boot loader to use UKI semantics (as exposed by Type #2) even though it is referenced via a Type #1
+  entry. This is useful to parameterize the UKI differently, to place it in a directory other than
+  `/EFI/Linux`, or to share a single UKI image among multiple menu entries.
+
+  If this key is set, and the system is not an EFI system, this entry should be hidden.
+
+  Example: `uki /fooos/bar.efi`
+
+* `uki-url` is similar to `uki`, but takes an RFC 3986 URI instead of a file path. It may be used to
+  reference remotely stored UKIs, which shall be downloaded using network boot support of the boot loader or
+  firmware on activation.
+
+  Example: `uki-url http://example.com/fooos.efi`
+
+  Optionally, in place of an URL a simple filename prefixed with `:` may be specified. In this case, if the
+  boot loader itself was booted via the network and knows its originating URL it shall replace the last
+  path component of that URL (i.e. the filename part) with the specified filename.
+
+  Example: `uki-url :fooos.efi`
+
+  This example would mean: if the boot loader was acquired from `http://example.com/somedir/fooosbootldr.img`
+  such a line would be treated equivalent to `uki-url http://example.com/somedir/fooos.efi`.
+
+  If this key is set, and the system is not an EFI system, or the boot loader/firmware does not support
+  network booting, this entry *shall* be hidden. Similarly, if the `:` syntax is used, but the boot loader
+  was not invoked via network booting (or doesn't know its originating URL) entries of this type *shall* be
+  hidden.
+
 * `options` shall contain kernel parameters to pass to the Linux kernel to
   spawn. This key is optional and may appear more than once in which case all
   specified parameters are combined in the order they are listed.
@@ -318,10 +348,10 @@ key. Here is an example for a complete drop-in file:
     linux        /6a9857a393724b7a981ebb5b8495b9ea/3.8.0-2.fc19.x86_64/linux
     initrd       /6a9857a393724b7a981ebb5b8495b9ea/3.8.0-2.fc19.x86_64/initrd
 
-On EFI systems all Linux kernel images should be EFI images. In order to
-increase compatibility with EFI systems it is highly recommended only to
-install EFI kernel images, even on non-EFI systems, if that's applicable and
-supported on the specific architecture.
+On EFI systems all Linux kernel images *should* be EFI PE images, or in case of the `uki` or `uki-url` key
+*must* be EFI PE images conforming to the [UKI Specification](unified_kernel_image.md). In order to increase
+compatibility with EFI systems it is highly recommended only to install EFI kernel images, even on non-EFI
+systems, if that's applicable and supported on the specific architecture.
 
 Conversely, in order to increase compatibility it is recommended to install
 generic kernel images that make few assumptions about the firmware they run on,


specs/boot_loader_specification.md
```diff
@@ -279,6 +279,36 @@ The following keys are recognized:
...
+  component of that URL (i.e. the filename part) with the specified filename.
+
+  Example: `uki-url :fooos.efi`
+
+  This example would mean: if the boot loader was acquired from `http://example.com/somedir/fooosbootldr.img`
+  such a line would be treated equivalent to `uki-url http://example.com/somedir/fooos.efi`.
+
+  If this key is set, and the system is not an EFI system, or the boot loader/firmware does not support
+  network booting, this entry should be hidden. Similar, if the `:` syntax is used, but the boot loader was
+  not invoked via network booting (or doesn't know its originating URL) entries of this key type shall be
```
keszybz, line 309: s/key type/type/. (Because "key type" is "uki-url", and "type" can be more specific.)

specs/boot_loader_specification.md
```diff
@@ -279,6 +279,36 @@ The following keys are recognized:
...
+
+  Example: `uki /fooos/bar.efi`
+
+* `uki-url` is similar to `uki`, but takes an RFC 3986 URI instead of a file path. It may be used to
+  reference remotely stored UKIs, which shall be downloaded using network boot support of the boot loader or
+  firmware on activation.
+
+  Example: `uki-url http://example.com/fooos.efi`
+
+  Optionally, in place of an URL a simple filename prefixed with `:` may be specified. In this case, if the
```
keszybz, line 298: Normally URLs work the way that a relative path is relative to the originating document, with no special prefix. This `:` is rather ugly… *If* we have a separate `uki-url` key, then we don't need specify `:`. In fact, it'd be better to follow normal HTTP rules, and allow `/some/path` to refer to the an absolute path on the same server, `foo/bar` to refer to a path relative to the bootloader (our "originating document"), and `http://some.address/full/url` for full URLs.

specs/boot_loader_specification.md
```diff
@@ -279,6 +279,36 @@ The following keys are recognized:
...
+  boot loader itself was booted via the network and knows its originating URL it shall replace the last
+  component of that URL (i.e. the filename part) with the specified filename.
```
behrmann, line 300: To maybe be strict here, since above the RFC is referenced:
```suggestion
  path component of that URL (i.e. the filename part) with the specified filename.
```

specs/boot_loader_specification.md
```diff
@@ -279,6 +279,36 @@ The following keys are recognized:
...
+  specified path shall refer to a [unified kernel image (UKI)](unified_kernel_image.md). This informs the
+  boot loader to use UKI semantics (as exposed by Type #2) even though it is referenced via a Type #1
+  entry. This is useful to parameterize the UKI differently, to place it in a directory other than
+  `/EFI/Linux`, or to share a single UKI image among multiple menu entries.
+
+  If this key is set, and the system is not an EFI system, this entry should be hidden.
+
+  Example: `uki /fooos/bar.efi`
+
+* `uki-url` is similar to `uki`, but takes an RFC 3986 URI instead of a file path. It may be used to
```
behrmann, line 292: Since the RFC is referenced here, I wonder, does it need both keys `uki` and `uki-url`? Since the URIs have a scheme, `uki-url file:///fooos/bar.efi` should have the same semantics as above, no? One could document that without a scheme prefixed `file://` is implied.

specs/boot_loader_specification.md
poettering: the only reason i did the `:` thing is that I am talking about `url` here, not `http`. So this could in theory be anything, i.e. nvmet:, iscsi: nfs: or ftp: or even file: or anyhting else even.

But if you read the rfc spec you'll find that they generically only say `<scheme>:<scheme specific part>` as syntax, which makes it really hard to distinguish simple file names from full urls... i.e. is "nvmet:192.168.1.1" a file or an url? it's not that obvious to me...

for web stuff, they know they know much better what to look for in an url than what we do, i.e. they know it's pretty much about http:// and https://, so they can look for that. but for boot stuff it would suck to build such a db of url prefixes, I doubt it's realistic.

hence i opted for the `:` prefix, because the URI RFC says the scheme must be `ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`, i.e. not `:`

does that make sense?

specs/boot_loader_specification.md
poettering: file:// is weird, it has a "host" anchor which noone knows what it's relative to. the type 1 stanzas such as linux/initrd/efi so far made very clear that they are relative to the ESP or the partition the snippets are placed in. you cannot really express that with file://.

the reason i split this up is mostly for the same ambiguation reasons described above, also i figured it's a good idea to distinguish "network boot" from "regular boot" with a very high-level construct, i.e. via a stanza. or to say this differently, if we'd shoehorn uri support into `uki`, then people would want it for `initrd` and `linux` and `efi` too, and I'd really like to avoid that. or with other words: i think it should be a property of an entry as a whole whether it is network based, not of the individual files listed therein, if that makes sense?

specs/boot_loader_specification.md
```diff
@@ -279,6 +279,36 @@ The following keys are recognized:
...
+
+  Example: `uki-url http://example.com/fooos.efi`
+
+  Optionally, in place of an URL a simple filename prefixed with `:` may be specified. In this case, if the
+  boot loader itself was booted via the network and knows its originating URL it shall replace the last
+  component of that URL (i.e. the filename part) with the specified filename.
+
+  Example: `uki-url :fooos.efi`
+
+  This example would mean: if the boot loader was acquired from `http://example.com/somedir/fooosbootldr.img`
```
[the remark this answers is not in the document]
poettering: uefi firmware boot theoretically supports https, but if you do you need to enroll the CAs for that too, i.e. the full CA bundle. I dn't think that is realistic in the global picture. THe CA bundle is huge, and I doubt firmwares are properly implementing hierarchies of CA... 

we don't really need https for resources of this type because of SecureBoot/shim authentication. it's fine if this is transferred unencrypted.

This discussion doesn't really fit into this spec, but I think what does fit into a spec is an exmaple that is actually realistic in the big picture.

specs/boot_loader_specification.md
behrmann: I'm fine with that reasoning.

My implicit assumption was, that `file://` in this context would anchor at the ESP like the other ones, since everything else would be confusing, i.e. seeing the firmware as a host on its own, with the ESP as its root.

When people would ask for network support for linux, initrd and friends, I'd politely point them to ipxe. :)

septatrix:
I think it would be helpful and reduce confusion if the design decisions mentioned in the above reviewed (uki vs uki-url, ":", and http/https) would be added as footnotes or in a closing section similar to the FAQ section of DDI and "Additional discussion" section of the version number spec

specs/boot_loader_specification.md
```diff
@@ -279,6 +279,35 @@ The following keys are recognized:
...
+  specified path shall refer to a [unified kernel image (UKI)](unified_kernel_image.md). This informs the
+  boot loader to use UKI semantics (as exposed by Type #2) even though it is referenced via a Type #1
+  entry. This is useful to parameterize the UKI differently, to place it in a directory other than
+  `/EFI/Linux`, or to share a single UKI image among multiple menu entries.
+
+  If this key is set, and the system is not an EFI system, this entry should be hidden.
+
+  Example: `uki /fooos/bar.efi`
+
+* `uki-url` is similar to `uki`, but takes an RFC 3986 URI instead of a file path. It may be used to
```
arianvp, line 292: IPXE supports specifiers in URLs e.g.

```
#!ipxe
initrd http://myserver.com/initrd?uuid=${uuid}&mac=${mac:hexhyp}
chain https://myserver.com/vmlinuz?uuid=${uuid}&mac=${mac:hexhyp}
```

which is great in datacenters as you can server specific images for specific servers dynamically.
e.g. with https://matchbox.psdn.io/matchbox/

Can we do something similar here?

arianvp: So a specifier for the machine-id and maybe the SMBIOS UUID (Though perhaps those are the same when SMBIOS UUID is available?)

arianvp: Fits in nicely with: https://github.com/systemd/systemd/pull/32086 because then the machine id and the SMBIOS UUID match

specs/boot_loader_specification.md
```diff
@@ -279,6 +279,36 @@ The following keys are recognized:
...
+  boot loader itself was booted via the network and knows its originating URL it shall replace the last
+  component of that URL (i.e. the filename part) with the specified filename.
+
+  Example: `uki-url :fooos.efi`
+
+  This example would mean: if the boot loader was acquired from `http://example.com/somedir/fooosbootldr.img`
+  such a line would be treated equivalent to `uki-url http://example.com/somedir/fooos.efi`.
+
+  If this key is set, and the system is not an EFI system, or the boot loader/firmware does not support
+  network booting, this entry should be hidden. Similar, if the `:` syntax is used, but the boot loader was
```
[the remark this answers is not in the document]
keszybz: "Similarly", and also we have "should" and "shall". Please use one or the other consistently, with `* *`.

keszybz (approved):
OK, I think we should merge this.

(I'm still not 100% convinced that `:` needs to be added, but I don't think it makes sense to bike-shed this.)

poettering:
I made the suggested changes, and will merge now, since they appear minor, and noone seems opposed.

PR #144, opened 2025-05-22
BLS: rely on link to UKI spec instead of redefining format 

bluca:
The only mandatory section is .linux. Rather than re-defining it, simply
rely on the UKI's own spec.


specs/boot_loader_specification.md
```diff
@@ -395,19 +395,6 @@ Images of this type have the advantage that all metadata and payload that makes
...
-* `.cmdline` section with the kernel command line,
-* `.osrel` section with an embedded copy of the
-  [os-release](https://www.freedesktop.org/software/systemd/man/os-release.html)
-  file describing the image.
-
-The `PRETTY_NAME=` and `VERSION_ID=` fields in the embedded `os-release` file
-are used the same as `title` and `version` in the Type #1 entries. The
-`.cmdline` section is used instead of the `options` field. `linux` and `initrd`
-fields are not necessary, and there is no counterpart for the `machine-id`
-field.
```
behrmann, line 409: Shouldn't this stay? The connection of these fields to type1 entries is not entirely clear from the UKI spec, e.g. for `.osrel` the UKI spec says
> They can be useful for presentation of the UKI in the boot loader menu, and ordering it against other entries using the included version information.

but the details are only here and I think would be the better place for them.

specs/boot_loader_specification.md
```diff
@@ -395,7 +395,8 @@ Images of this type have the advantage that all metadata and payload that makes
 up the boot entry is contained in a single PE file that can be signed
 cryptographically as one for the purpose of EFI SecureBoot.
 
-A valid unified kernel image in the `/EFI/Linux/` directory must contain two PE sections:
+A unified kernel image in the `/EFI/Linux/` directory may contain the following
+PE sections:
```
behrmann, line 399: I think it would make sense to drop the `.cmdline` section from the list below (and not make this a list at all then), since there may be many other sections. How about something something likes this

```
A unified kernel image in the `/EFI/Linux/` directory contains PE sections described in the [UKI Specification](unified_kernel_image.md).

Of these sections, the `.osrel` section is used to generate boot entries. 
The `PRETTY_NAME=` and `VERSION_ID=` fields in the embedded `os-release` file
are used the same as `title` and `version` in the Type #1 entries. The
`.cmdline` section is used instead of the `options` field. `linux` and `initrd`
fields are not necessary, and there is no counterpart for the `machine-id`
field.
```

PR #145, opened 2025-08-07
specs: port file-hierarchy specification from systemd manpages

bluca:
https://www.freedesktop.org/software/systemd/man/latest/file-hierarchy.html

Converted with pandoc from docbook and then manually reformatted

### DIFF
--- a/specs/filesystem_hierarchy.md
+++ b/specs/filesystem_hierarchy.md
@@ -1,0 +1,592 @@
+---
+title: FileSystem Hierarchy
+category: Concepts
+layout: default
+version: 1
+SPDX-License-Identifier: CC-BY-4.0
+---
+
+# FileSystem Hierarchy
+
+## Description
+
+Operating systems using the
+[`systemd(1)`](https://www.freedesktop.org/software/systemd/man/systemd.html)
+system and service manager are organized based on a file system hierarchy inspired by UNIX,
+more specifically the hierarchy described in the
+[File System Hierarchy](http://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html)
+specification and
+[`hier(7)`](https://man7.org/linux/man-pages/man7/hier.7.html),
+with various extensions,
+partially documented in the
+[XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+and [XDG User Directories](https://www.freedesktop.org/wiki/Software/xdg-user-dirs).
+This manual page describes a more generalized, though minimal and modernized
+subset of these specifications that defines more strictly the suggestions and
+restrictions systemd makes on the file system hierarchy.
+Note that this document makes no attempt to define the directory structure comprehensively,
+it only documents a skeleton of a directory tree, that downstreams can extend.
+Because of that traditional directories such as
+`/usr/include/`
+or
+`/var/spool/`
+are not covered,
+even though it might (or might not) make a lot of sense to include them
+in the structure of an actually deployed OS.
+
+Many of the paths described here can be queried with the
+[`systemd-path(1)`](https://www.freedesktop.org/software/systemd/man/systemd-path.html)
+tool.
+
+## General Structure
+
+### `/`
+
+The file system root.
+Usually writable, but this is not required.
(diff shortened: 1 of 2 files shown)


specs/filesystem_hierarchy.md
```diff
@@ -0,0 +1,529 @@
...
+system and service manager are organized based on a file system hierarchy
+inspired by UNIX, more specifically the hierarchy described in the [File System
+Hierarchy](http://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html)
+specification and
+[`hier(7)`](https://man7.org/linux/man-pages/man7/hier.7.html), with various
+extensions, partially documented in the [XDG Base Directory
+Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+and [XDG User
+Directories](https://www.freedesktop.org/wiki/Software/xdg-user-dirs). This
+manual page describes a more generalized, though minimal and modernized subset
```
keszybz, line 24: Not "manual page" now. Actually, that whole paragraph needs to be rewritten. So let's merge this as is, i.e. with unchanged text, and then update later.

specs/filesystem_hierarchy.md
```diff
@@ -0,0 +1,529 @@
...
+
+### `/boot/`
+
+The boot partition used for bringing up the system. On EFI systems, this is
+possibly the EFI System Partition (ESP), also see
+[`systemd-gpt-auto-generator(8)`](https://www.freedesktop.org/software/systemd/man/systemd-gpt-auto-generator.html).
+This directory is usually strictly local to the host, and should be considered
+read-only, except when a new kernel or boot loader is installed. This directory
+only exists on systems that run on physical or emulated hardware that requires
+boot loaders.
```
keszybz, line 52: "This directory only exists" — the intent was good, but I think the wording could be improved. We'd often create the directory generically, and just not use it on systems that don't need it.

specs/filesystem_hierarchy.md
```diff
@@ -0,0 +1,529 @@
...
+
+### `/efi/`
+
+If the boot partition `/boot/` is maintained separately from the EFI System
+Partition (ESP), the latter is mounted here. Tools that need to operate on the
+EFI system partition should look for it at this mount point first, and fall back
+to `/boot/` — if the former does not qualify (for example if it is not a mount
```
keszybz, line 59: This maze of "former" and "latter" makes this hard to read…

specs/filesystem_hierarchy.md
```diff
@@ -0,0 +1,529 @@
...
+point or does not have the correct file system type `MSDOS_SUPER_MAGIC`).
+
+### `/etc/`
+
+System-specific configuration. This directory may or may not be read-only.
```
keszybz, line 64: "may or may not" is usually spelled "may" :)

specs/filesystem_hierarchy.md
```diff
@@ -0,0 +1,529 @@
...
+
+For unprivileged system processes, only `/tmp/`, `/var/tmp/` and `/dev/shm/` are
+writable. If an unprivileged system process needs a private writable directory
+in `/var/` or `/run/`, it is recommended to either create it before dropping
+privileges in the daemon code, to create it via
+[`tmpfiles.d(5)`](https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html)
+fragments during boot, or via the `StateDirectory=` and `RuntimeDirectory=`
+directives of service units (see
+[`systemd.unit(5)`](https://www.freedesktop.org/software/systemd/man/systemd.unit.html)
+for details).
```
keszybz, line 421: I think we'd want to change the order here. Creation of the directory in daemon code is probably not the best option, esp. if it's the primary reason to *have* the privileges in the first place.

keszybz (commented):
Thanks. I think we want to merge this without semantic changes, and then maybe do some cleanups to the text later. I made some comments for later about that.

I think we should use semantic line breaks though. It seems that we'll be making changes to this text, and it'd be much nicer for the diffs to be meaningful.

bluca:
> I think we should use semantic line breaks though.

I am still not sure I know what that means, in practice, though :-)

PR #148, opened 2025-08-08
LFSH: drop mention of expectations of systemd from intro

bluca:
This is meant to be generic and neutral, so drop the mention from the intro

### DIFF
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -10,10 +10,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 ## Description
 
-This page describes the layout of a modern Linux system,
-as expected by the
-[`systemd(1)`](https://www.freedesktop.org/software/systemd/man/systemd.html)
-system and service manager.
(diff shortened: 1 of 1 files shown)


bluca:
Various systemd tools and things are mentioned through the doc, and that's fine for the reasons you said. But I think it would just look better, and stop people from instantly dismissing this, by not having this bit in the _intro_ itself, as the first line of the first paragraph.

It's also factually incorrect - even certain luddite distros that shall not be named, that won't ever get anywhere close to anything related to systemd, implement this hierarchy, including usr-merge. I think this should describe the situation at large, and changing the first line helps with it.

I'm not going to suggest editing any other line/paragraph for this, I am merely interested in the optics of the first line in the first paragraph, that's all.

keszybz:
OK, let's merge this then. I think we'll need to add a "See also" section with some links anyway, so we can link to systemd there.

bluca:
Yes, that sounds like a good idea, I'll send a follow-up to do that

bluca:
--> https://github.com/uapi-group/specifications/pull/151

PR #149, opened 2025-08-08
don't suggest people to use /usr/libexec/

poettering:


keszybz:
I guess the approach to this depends on how we want this document to function. When it was a man page in the systemd project, it was fine to be a bit terse and only describe some things in passing. But we moved it to the new location to make it more independent. If we want it to function on its own, then it needs to be comprehensive. In particular, we didn't need to describe `/usr/include` because systemd doesn't care about it, but if the doc is supposed to describe a "linux systems", like the title says, then I think we'll need to do that. Otherwise, the text is not stand-alone and requires people to link to other docs to too.

The situation with `/usr/libexec` is similar. If we describe it, even with a single short paragraph, then the topic can be closed. And I think this is very good to close the topic. I know you think the directory is not needed at all, which agree with, but OTOH, it is not harmful. There is just so much mostly futile discussions whether to use the directory or not… Some distros mandate it, others forbid it, and if we say it *may* be used, this document is acceptable to both. I was very happy to trim down the discussion of `libexec` in Fedora Packaging Guidelines, but that is only possible if we actually briefly describe it here.

I you think `/usr/libexec` is better avoided, then I'd prefer to keep the existing separate section and change the wording to recommend the other approach.

bluca:
And yes, I think we should describe /usr/include, I'll prep that separately

bluca:
> And yes, I think we should describe /usr/include, I'll prep that separately

---> https://github.com/uapi-group/specifications/pull/152

poettering:
I think it *is* harmful, because it complicates stuff for no reason...

poettering:
i mean, let's figure out what this document should document. The status quo? Or the ideal scenario? My goal with my original text was the "ideal scenario, but explain how the status quo can be evolved into it". 

If this spec is supposed to reain that goal, then libexec should not be named prominently as something to support, but only as a compat thing, or example for extending the skeleton or so.

bluca:
> i mean, let's figure out what this document should document. The status quo? Or the ideal scenario? My goal with my original text was the "ideal scenario, but explain how the status quo can be evolved into it".
> 
> If this spec is supposed to reain that goal, then libexec should not be named prominently as something to support, but only as a compat thing, or example for extending the skeleton or so.

I think that was fine for our own manpages (and we can keep them separate), but for _this_ one, in _this_ version, I think the best approach is to go for "accepted and widespread status quo" for version 1.
If we push out something that says "by the way 99% of you are non-compliant, lol", we are not going to make a lot of friends.

So we might soften the wording to make it sound like it's not mandatory and maybe not even recommended, but it exists and has its own section to accurately describe it.

And then perhaps we can plan for the next version, where things we don't want anymore for the future are hard deprecated?

We need to figure out how to handle versioning for everything here btw, right now the website just shows a single instance of the md doc, we might want more advanced stuff like drop-down selector for a version or stuff like that

bluca:
The thing is, for something to be called 'legacy' it is not enough that some of us don't like it. I don't like debian-style multiarch libdirs, and yet I'm not calling it legacy and I'm not asking for it to be dropped from the spec.

Things like non-merged-usr layouts are legacy because debates happened, they were settled, and the vast, vast majority has moved away from it.

Has there been a movement to remove libexec? Has the majority or even just a substantial number removed it? Is there a sustained push to get away from it?

I really never followed this topic, and again I have zero opinions on it. But my impression is that the answer to these questions *right now* is *no*. Which doesn't mean it's a good thing, or that I like it, or that I endorse it, it just is.

So yeah I am pretty sure in *v1* it should be defined. Which doesn't mean strongly endorsed nor recommended, but merely described.

Why can't this removal wait for a *v2*? Don't you think you'd have better chances to convince others to drop it if the starting point was that *v1* of this spec is adopted as the source of truth, and then starting a debate about *v2*, rather than going all-or-nothing from day 1?

keszybz:
Alternative approach in https://github.com/uapi-group/specifications/pull/155.

lnussel:
FWIW libexec appeared in (open)SUSE only recently by basically giving up for the sake of compatibility rather than acknowledging that it makes sense. For decades packages were supposed to use subdirs of `/usr/lib` :
https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/NKGQYNSEI4EFIHXHAKOBOISADFAHUNVW/
So if the intention of this specification is to describe the ideal layout for an OS from scratch, then libexec could certainly be omitted IMHO. OTOH I'd also question the necessity of some other locations too then :)

keszybz:
https://github.com/uapi-group/specifications/pull/155 was merged. This is obsolete now.

PR #150, opened 2025-08-08
Revert "LFSH: tone down text about shared ownership in /usr/share"

poettering:
This reverts commit 1f621b31e389c0d0519f1035a4e1e0fecf9617e4.

The reverted commit suggests that tmpfiles.d/ was a great example of shared ownership on equal footing, but that's simply not true, it's quite the contrary: the dir is owned a lot more by systemd-tmpfiles than by the packages dropping in files there, because the primary (and sole) consumer of those files is systemd-tmpfiles and – most importantly – the format is defined by systemd-tmpfiles and controlled by it, and by nothing else.

This creates an "imbalance of power" if you so will: one (singular) side defines and consumes, the other sides (plural) provide. And this is definitely not a "shared ownership on equal footing", it's quite the opposite.

And this to me is the fundamental difference that /usr/share/ vs /usr/lib/ really should be: resources in one are owned/controlled by some project in a fundamental way the other contains inherently shared resources, where *everyone* is on equal footing.

### DIFF
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -223,9 +223,11 @@ such as documentation, man pages, time zone information, fonts and other resourc
 Usually, the precise location and format of files stored below this directory
 is subject to specifications that ensure interoperability.
 
-Note that resources placed in this directory may be under shared ownership,
-with multiple different packages providing and consuming these resources
-on equal footing without any obvious primary owner.
+Note that resources placed in this directory typically are under shared ownership,
+i.e. multiple different packages have provided and consumed these resources,
+on equal footing, without any obvious primary owner.
+This makes things systematically different from `/usr/lib/`,
+where ownership is generally not shared.
 
 ### `/usr/share/doc/`
 


specs/linux_file_system_hierarchy.md
```diff
@@ -223,9 +223,11 @@ such as documentation, man pages, time zone information, fonts and other resourc
 Usually, the precise location and format of files stored below this directory
 is subject to specifications that ensure interoperability.
 
-Note that resources placed in this directory may be under shared ownership,
-with multiple different packages providing and consuming these resources
-on equal footing without any obvious primary owner.
+Note that resources placed in this directory typically are under shared ownership,
```
bluca, line 226: "are typically"

specs/linux_file_system_hierarchy.md
```diff
@@ -223,9 +223,11 @@ such as documentation, man pages, time zone information, fonts and other resourc
...
+i.e. multiple different packages have provided and consumed these resources,
```
bluca, line 227: `i.e.:` or `i.e.,` would look nicer to me

keszybz:
I'll copy the comment I made in #147:

> Sorry, but that just doesn't match reality. The difference between `/usr/share` and `/usr/lib` is that one is mostly used for arch-indepedent stuff, and the other not. But in general, any location that is used cooperatively by multiple components requires … cooperation, a shared understanding of how things should be named to avoid conflicts and what formats are allowed. In the terms you describe, `/usr/share/man` is "owned" by man, and we have a very good understanding of how files must be named there and what formats are allowed. The packaging guidelines even specify what compression must be used and the file glob format in package specs.  The same for `/usr/share/pkgconfig/`. The text that was removed created a distinction that isn't followed in practice and (at least to me) doesn't make much sense.

> Looking at my system, I have hundreds of directories under `/usr/share`, and the big majority is just the typical case of something used by a single package or a group of closely-cooperating packages. The removed text said "resources placed in this directory *typically* are under shared ownership, i.e. multiple different packages have provided and consumed these resources, on equal footing, without any obvious primary owner", which doesn't seem to be the case at all.

keszybz:
OK, there have been no further comments. Since this seems to have been just a misunderstanding, I'll close this.

PR #151, opened 2025-08-08
LFSH: add 'See Also' section and specifically mention systemd and link it

bluca:


keszybz (approved):
Yeah, that is nice.

poettering:
The reference to systemd is a bit weird given that systemd neither implements the libexec thing nor the new ownership rules placed here. Maybe drop that reference here because it follows a different model?

Sorry to be the party pooper here, but if you make recommendations that systemd doesnt actually follow and really shouldnt then systemd really shouldnt act as a poster child, because this here is something else that might be inspired by systemd but defines distinct rules systemd.

PR #152, opened 2025-08-08
LFHS: describe '/usr/include/'

bluca:


keszybz:
> `/usr/lib/arch-id/` has a separate section, but here I added `/usr/include/arch-id/` to the same as `/usr/include/`. Not sure which one is better?

Yeah, I think that's good. It's a short text anyway.

specs/linux_file_system_hierarchy.md
```diff
@@ -176,6 +176,18 @@ It is recommended not to place binaries in this directory
...
+Programmatic APIs header files, typically used when compiling `C/C++` programs.
+Packages may place such files either directly in `/usr/include/`,
+or in a subdirectory.
+The subdirectory may be named after the package or project providing it.
+Packages may place architecture-dependent header files and directories in a
+`/usr/include/arch-id/` subdirectory,
+following the identifiers defined on the
+[Multiarch Architecture Specifiers (Tuples)](https://wiki.debian.org/Multiarch/Tuples)
+list.
+
```
keszybz, line 190: I had the following additions in my version:
```
The corresponding library shall be placed in `$libdir`, see below.
```
and
```
Using a subdirectory is recommended if the package installs multiple files.
It is also necessary if multiple versions of the same headers shall be coinstallable.
```
ad
```
Note that to use the headers, flags for the compiler might be required
to bring the appropriate subdirectory into the search path.
See man pages
[`pc(7)`](https://www.mankier.com/5/pc),
[`pkgconf(1)`](https://www.mankier.com/1/pkgconf),
[gcc's Search Path](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html),
and other compiler specific documentation.
```

The mention of `/usr/include` also needs to be dropped from the intro where we say we don't document it.

bluca: added

specs/linux_file_system_hierarchy.md
```diff
@@ -176,6 +172,30 @@ It is recommended not to place binaries in this directory
...
+Packages may place such files either directly in `/usr/include/`,
+or in a subdirectory.
+Using a subdirectory is recommended if the package installs multiple files.
+It is also necessary if multiple versions of the same headers shall be coinstallable.
+The subdirectory may be named after the package or project providing it.
+Packages may place architecture-dependent header files and directories in a
+`/usr/include/arch-id/` subdirectory,
+following the identifiers defined on the
+[Multiarch Architecture Specifiers (Tuples)](https://wiki.debian.org/Multiarch/Tuples)
+list.
```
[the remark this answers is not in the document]
bluca: It is supported and has been since the moment /usr/lib/arch-id/ was in distributions that do multi-arch:

```
$ apt-file search /usr/include/x86_64-linux-gnu | wc -l
30474
```

poettering:
so what is this document supposed to be again? list an "ideal scenario"? then why would it include /usr/include? list "status quo"? then why list this weird, made-up per-arch include dir?

i don't think any of this comes with enough careful thinking? figure out what you want to do, figure out why you want to do it, then document the result, but this just appears totally random to me, with no real thought given (how could it, most of the PRs prepped for it, barely had the chance to be reviewed, where merged after minutes, not hours of permitting reviews)

the original doc in the man page is relatively carefully put together, there's a reason for everything listed there. might not be perfect, but i can defend every item, and systemd implements every item listed. it's an "ideal" thing, but mentions some common extensions briefly. I think that's very useful, but this doc here doesn't make clear at all what it is, claims systemd would implement it, though it really doesn't, and just muddies the waters, since people assume this was widely accepted, but really isn't, not even inside the systemd maintainer team.

bluca:
> so what is this document supposed to be again? list an "ideal scenario"? then why would it include /usr/include? list "status quo"? then why list this weird, made-up per-arch include dir?

It's status quo, and as mentioned above, per-arch include dir exists everywhere per-arch libdir exists, since the very beginning

PR #155, opened 2025-08-08
LFSH: say that libexec is not recommended

keszybz:
libexec is a very heated topic. There is no particular technical rationale for this directory, because a subdirectory under /usr/lib is generally equivalent. So we shouldn't *recommend* the use of /usr/libexec. Nevertheless, it is fairly widely used by many distributions (Fedora, RHEL, SUSE, Mandriva, now also Debian), so we want to describe this directory so people know how it is used when they see it on their systems. It is now described as "secondary" and "alternative". I didn't want to use the word "legacy", because there is no plan to get rid of it in the foreseeable future.

Altenative to https://github.com/uapi-group/specifications/pull/149.

### DIFF
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -34,6 +34,10 @@ Some directories like `/var/spool/` are not covered,
 even though it might make sense to include them
 in the structure of an actually deployed OS.
 
+Some directories are described for compatibility with current Linux distributions,
+but their use is not recommended.
+The subsections that describe those directories are marked with ⚠️.
+
 Many of the paths described here can be queried with the
 [`systemd-path(1)`](https://www.freedesktop.org/software/systemd/man/systemd-path.html)
 tool, on systems where this tool is available.
@@ -81,8 +85,8 @@ following the
 [Configuration Files Specification](configuration_files_specification.md).
 
 If `/opt/` is supported (see below),
-then the `/etc/opt/` subdirectory is the location where third-party software installed in `/opt/` may store
-their configuration.
+then the `/etc/opt/` ⚠️ subdirectory is the location where
+third-party software installed in `/opt/` may store its configuration.
 The same naming convention as directories under `/opt/` is used for directories under `/etc/opt/`.
 
 ### `/home/`
@@ -98,7 +102,7 @@ Applications should generally not reference this directory directly,
 but via the per-user `$HOME` environment variable,
 or via the home directory field of the user database.
 
-### `/opt/`
+### `/opt/` ⚠️
 
 A secondary location for third-party vendor directories.
 This directory is optional,
@@ -235,7 +239,7 @@ Such binaries may be for any architecture supported by the system.
 Do not place public libraries in this directory,
 use `$libdir` (see below), instead.
 
-### `/usr/libexec/`
+### `/usr/libexec/` ⚠️
 
 A secondary location for
 vendor binaries or other programs that are not regularly invoked from a shell
@@ -330,10 +334,11 @@ though it is recommended to do most logging via the
 [`sd_journal_print(3)`](https://www.freedesktop.org/software/systemd/man/sd_journal_print.html)
 calls.
 
-### `/var/opt/`
+### `/var/opt/` ⚠️
+
 If `/opt/` is supported (see above),
-then the `/var/opt/` subdirectory is the location where third-party software installed in `/opt/` stores its
-persistent, variable data.
+then the `/var/opt/` subdirectory is the location where third-party software installed in `/opt/`
+stores its persistent, variable data.
 The same naming convention as directories under `/opt/` is used for directories under `/var/opt/`.
 
 ### `/var/tmp/`


bluca (approved):

specs/linux_file_system_hierarchy.md
```diff
@@ -210,16 +210,18 @@ use `$libdir` (see below), instead.
 
 ### `/usr/libexec/`
 
-Vendor binaries or other programs that are not regularly invoked from a shell.
+An secondary location for
```
behrmann, line 213:
```suggestion
A secondary location for
```

cgwalters (approved):

keszybz:
Based on the discussion during the meeting, I updated the wording to say that `libexec` is not recommended more strongly, and added a similar commit for `/opt/`, and some icons. @poettering PTAL.

bluca:
For libexec it is fine, but the reason opt exists is that often third party vendors bundle their own builds of shared libraries like openssl, and we really, really don't want them to clobber the os vendors files. That's why there's a separate hierarchy for third party owned stuff, and given there's no alternative (unlike libexec, where there is), I'd remove the "not recommended" wording. The optional wording is fine of course

specs/linux_file_system_hierarchy.md
```diff
@@ -231,16 +231,19 @@ use `$libdir` (see below), instead.
 
 ### `/usr/libexec/`
```
poettering: hmm, what about adding some qualifier here to the section title like `(facultative)` or `(optional)` or `(dispensable)` or so? That would mean it would already in the TOC be visible as optional

specs/linux_file_system_hierarchy.md
```diff
@@ -100,13 +100,19 @@ or via the home directory field of the user database.
...
-Third-party vendors (i.e.: unrelated to the OS provider) use a directory,
+A secondary location for third-party vendor directories.
+This directory is optional,
+as not all systems allow installing third-party software
+or using `/opt/` even if they do.
+Using the normal hierarchy under `/usr/`,
+integrated with the rest of the distribution,
+is recommended instead.
+
+Each third-party vendor (i.e.: unrelated to the OS provider) may use a subdirectory,
```
poettering, line 115: a *private* subdirectory

specs/linux_file_system_hierarchy.md
```diff
@@ -168,10 +172,11 @@ Runtime system logs.
 System components may place private logs in this directory.
 Always writable, even when `/var/log/` might not be accessible yet.
 
-### `/run/opt/`
+### `/run/opt/` ⚠️
```
poettering, line 175: what dir is this? this is completely made up, i cannot find any reference to it on the internet. 

It's one thing to be descriptive of stuff that is unnecessary/broken/bad but has tradition. but this dir seems like it is made up by the spec, and we shouldn't add new stuff to those legacy parts really.

bluca: It's the corresponding of /etc/opt and /var/opt, but for runtime data, some 3rd party sw from microsoft uses it

bluca: and /opt is optional, not legacy

poettering:
> For libexec it is fine, but the reason opt exists is that often third party vendors bundle their own builds of shared libraries like openssl, and we really, really don't want them to clobber the os vendors files. That's why there's a separate hierarchy for third party owned stuff, and given there's no alternative (unlike libexec, where there is), I'd remove the "not recommended" wording. The optional wording is fine of course

/opt/ doesn't work that way. Because nothing included shows up in any search paths. Not in $PATH, not in $LD_LIBRARY_PATH, not in the XDG search paths, or .desktop search paths, or icon search paths, or mime registrarion search paths, or systemd unit file search paths or anything else. Hence, any real-life software using /opt/ places stuff all over /usr/ anyway, to get their stuff included in the search paths. But that means /opt/ doesn't get you anything, because the dreamt up separation of os vendor and app vendor resources cannot be achieved this way.

But they need to place stuff into /usr/ anyway, then they might as well place whatever they'd put in `/opt/<project>` into `/usr/lib/<project>/` and avoid the hassle.

that's the thing with /opt/, its completely underdefined and hence *not* a recommended thing.

I think /usr/local/ is better defined than /opt/, because it typically *is* included in search paths.

Hence, I am strongly for keeping the wording that /opt/ ain't recommended.

bluca:
> > For libexec it is fine, but the reason opt exists is that often third party vendors bundle their own builds of shared libraries like openssl, and we really, really don't want them to clobber the os vendors files. That's why there's a separate hierarchy for third party owned stuff, and given there's no alternative (unlike libexec, where there is), I'd remove the "not recommended" wording. The optional wording is fine of course
> 
> /opt/ doesn't work that way. Because nothing included shows up in any search paths. Not in $PATH, not in $LD_LIBRARY_PATH, not in the XDG search paths, or .desktop search paths, or icon search paths, or mime registrarion search paths, or systemd unit file search paths or anything else. Hence, any real-life software using /opt/ places stuff all over /usr/ anyway, to get their stuff included in the search paths. But that means /opt/ doesn't get you anything, because the dreamt up separation of os vendor and app vendor resources cannot be achieved this way.
> 
> But they need to place stuff into /usr/ anyway, then they might as well place whatever they'd put in `/opt/<project>` into `/usr/lib/<project>/` and avoid the hassle.
> 
> that's the thing with /opt/, its completely underdefined and hence _not_ a recommended thing.
> 
> I think /usr/local/ is better defined than /opt/, because it typically _is_ included in search paths.
> 
> Hence, I am strongly for keeping the wording that /opt/ ain't recommended.

The files other application owns, like desktop files or unit files, are placed where the apps that own them decide to. And it's fine, because they are namespaced. LD_LIBRARY_PATH etc do not matter, as those are all private things, and are built with RPATH and so on, so it's in fact _good_ that they don't show up in paths - they are not supposed to, and it would be a breakage if it did. My OS provided app should most definitely not use by accident libEGL provided by chrome.

Of course if you want to start providing paths where third party vendors can drop unit files, that is separate from the OS one, that would be fine.

> nah, they wouldn't, they could put that stuff into /usr/lib/google/chrome/ or /usr/lib/Signal or /usr/lib/keybase and be just like any other package, and without conflicting more than they already conflict.

But they don't, and won't, so it's moot. /opt/ is where they are installed, and where they will continue to install, and it's good as usually these packages are poor quality, so nobody trusts them not to install libraries in /usr/lib/, so it's good they use --prefix=/opt when building. I understand you don't like it, but that doesn't automatically mean it's legacy or deprecated. It's optional, and that's enough. This spec needs to describe the world as it is and as it's going to be, otherwise it's useless.

specs/linux_file_system_hierarchy.md
keszybz: See the third commit.

keszybz:
> I think /usr/local/ is better defined than /opt/, because it typically _is_ included in search paths.

Please don't mix `/usr/local` into this. It is used by local builds and such, and should not be used by vendor packages.

poettering:
> The files other application owns, like desktop files or unit files, are placed where the apps that own them decide to. 

desktop files are not owned by anyone really except the app installing them. they are a generic spec implemented by many things.

> And it's fine, because they are namespaced. LD_LIBRARY_PATH etc do not matter, as those are all private things, and are built with RPATH and so on, so it's in fact good that they don't show up in paths - they are not supposed to, and it would be a breakage if it did. My OS provided app should most definitely not use by accident libEGL provided by chrome.

There's no conflict if they put stuff in /usr/lib/<package>.

> nah, they wouldn't, they could put that stuff into /usr/lib/google/chrome/ or /usr/lib/Signal or /usr/lib/keybase and be just like any other package, and without conflicting more than they already conflict.

But they don't, and won't, so it's moot. /opt/ is where they are installed, and where they will continue to install, and it's good as usually these packages are poor quality, so nobody trusts them not to install libraries in /usr/lib/, so it's good they use --prefix=/opt when building. I understand you don't like it, but that doesn't automatically mean it's legacy or deprecated. It's optional, and that's enough. This spec needs to describe the world as it is and as it's going to be, otherwise it's useless.

man, it such *pain* arguing with you on these things. can't you at least accept that there's no agreement on this, and hence the thing should not be recommended like the others are?

specs/linux_file_system_hierarchy.md
poettering: yeah, i guess the next sentence clears that up. it's fine to ignore this hence

poettering:
just to mention: i think the spec is fine with this applied + the reference to /run/opt/ removed.

keszybz:
I added one more commit to drop the mention of `/run/opt`.

bluca:
> > The files other application owns, like desktop files or unit files, are placed where the apps that own them decide to.
> 
> desktop files are not owned by anyone really except the app installing them. they are a generic spec implemented by many things.
> 
> > And it's fine, because they are namespaced. LD_LIBRARY_PATH etc do not matter, as those are all private things, and are built with RPATH and so on, so it's in fact good that they don't show up in paths - they are not supposed to, and it would be a breakage if it did. My OS provided app should most definitely not use by accident libEGL provided by chrome.
> 
> There's no conflict if they put stuff in `/usr/lib/<package>`.

There is, as OS vendor may ship the same package too. There was recently a thread by someone working to package signal natively for debian for example.

> > > nah, they wouldn't, they could put that stuff into /usr/lib/google/chrome/ or /usr/lib/Signal or /usr/lib/keybase and be just like any other package, and without conflicting more than they already conflict.
> > 
> > 
> > But they don't, and won't, so it's moot. /opt/ is where they are installed, and where they will continue to install, and it's good as usually these packages are poor quality, so nobody trusts them not to install libraries in /usr/lib/, so it's good they use --prefix=/opt when building. I understand you don't like it, but that doesn't automatically mean it's legacy or deprecated. It's optional, and that's enough. This spec needs to describe the world as it is and as it's going to be, otherwise it's useless.
> 
> man, it such _pain_ arguing with you on these things. can't you at least accept that there's no agreement on this, and hence the thing should not be recommended like the others are?

The feeling is mutual. Just because you like or prefer something, it doesn't automatically mean it trumps reality. The point of this doc is to recognize and describe the world as-is, as current software uses it, not as one individual would like it to be for their own personal preferences, against everything else. All these things exist, they exist for a variety of reasons, even if anyone disagree with those reasons, and they are not going to magically go away.

If you want to mark /opt/ as deprecated, then _first_ you need to go out there and do the leg work to make it happen. Just like merged-sbin, it can be marked as status quo not because someone somewhere said they like it, but because they did _work_ to make it happen. The doc follows reality, not the other way around.

This is not a wishlist, and that's good, because it would not really be useful as a wishlist. There are other places where wishlists can be recorded.

So if you can go to Google/Microsoft/Signal/etc, all the popular third party stuff that ships in /opt/, and you can make them change to your preferred layout, _then_ we can change the doc to match the new reality.

Until that happens, when a Fedora user or Ubuntu user looks at their system, and look at this spec, and see that /opt/ is there, they should not get the impression that there is something wrong with their installation, that it's somehow broken.

bluca:
> > If you want to mark /opt/ as deprecated, then _first_ you need to go out there and do the leg work to make it happen. Just like merged-sbin, it can be marked as status quo not because someone somewhere said they like it, but because they did _work_ to make it happen. The doc follows reality, not the other way around.
> 
> I don't think that this description strikes the correct balance. I agree that we need to describe `/opt/`, because it _is_ widely used. But that doesn't mean that we can't recommend using a different approach. Even when we do that, we are not putting any due date for the removal, and vendors can continue doing what they are doing.

I don't think the current wording quite hits the right spot either though:

> Using `/opt/` is not recommended.
> It is not integrated with the rest of the distribution:
> a package which uses `/opt/` would anyway need to install
> binaries or links in `/usr/bin/`
> and desktops files in one of the appropriate locations.
> Instead of using a subdirectory under `/opt/`,
> a third party vendor should put their directory under `/usr/lib/`.

The first sentence might be ok as a compromise. But the rest seems very out of place, as it attempts to find some justifications that really don't match reality. These are not real problems, there are solutions for them and they are currently used, so it looks very strange and weird, as if it the document was going out of its way to willingly ignore things, and justify itself.

Just drop everything but the first line?

> > There is, as OS vendor may ship the same package too. There was recently a thread by someone working to package signal natively for debian for example.
> 
> Yes, a conflict would happen in that case. But actually, it would happen anyway, for the names of binaries and desktop files and man pages and so on. And even by some chance the two packages avoid file-level conflict, the user would most likely be confused by seeing two instances of the application. So in reality the conflict must be solved by either not installing the two copies at the same time or by renaming one of the versions so that it is distinguishable from the other one.

But this doesn't actually happen. These things exists and ship, and have for a long time, and we can't just ignore it.

This goes back to your original point: this doc needs to describe reality and be useful to end users as well. If a Fedora user installs Signal and then sees this document, what are they going to think? That they made a mistake and they need to correct it? That their installation is now broken and that there will be conflicts, causing errors?

poettering:
LGTM

bluca:
> I don't understand. What do you "don't match reality"? `rpm -ql google-chrome-stable` shows that those workarounds are actually used. The problem obviously exists and is real, that is why the workarounds are needed. And since those workarounds are needed, the very idea of `/opt/` is only half-baked.

What I mean is that we can call them workarounds, or hacks, or mechanisms, or any other negative/positive/neutral opinion, but they exist and are used effectively. The example you used, chrome, uses update-alternatives for example (maybe not the rpm? the deb does). Pretty or ugly as it may be, it works and it is used. There are no conflicts in the sense that nothing breaks just because it is installed, so we shouldn't suggest it does, because it's just not true

> I decided to add this justification in the text because this is bound to come up again and again.

I don't know how to describe, reading it just feels.. out of place? Awkward? It looks like a text that it's desperately trying to justify itself, but can't? Something like that. It feels like something that should be in a commit message, or a non-rendered comment if such a thing exists for markdowns, but not in the section?

poettering:
One thing I'd like to mention: The section at the bottom of the document makes very clear recommendations for package maintainers where to put stuff, once in user and once in system context. This does not mention /opt/ at all, because all these reasons: it's not being thought out, and also: quite redundant in my PoV.

If the language regarding /opt/ doesn't make clear this is not the preferred place to put stuff you just muddy the water: different parts of the spec suggest different things, leaving readers at a loss what they should actually follow: place stuff in `/opt/<package>/` or in `/usr/lib/<package>/`

specs/linux_file_system_hierarchy.md
```diff
@@ -98,15 +102,25 @@ Applications should generally not reference this directory directly,
...
 It is usually read-only, but this is not required.
 This directory should not be modified by the administrator,
 except when installing or removing third-party-supplied software.
-This directory is optional, as not all systems allow installing third-party software.
+
+Using `/opt/` is not recommended.
+It is not integrated with the rest of the distribution:
+a package which uses `/opt/` would anyway need to install
+binaries or links in `/usr/bin/`
+and desktops files in one of the appropriate locations.
```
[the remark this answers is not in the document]
bluca: Even that is still incorrect: a package doesn't need to install anything under /usr/bin, nor manual pages. That's what I mean: it's trying to justify itself by coming up with a strawman argument and trying to refute it, but it really doesn't work

keszybz:
> What I mean is that we can call them workarounds, or hacks, or mechanisms, or any other negative/positive/neutral opinion, but they exist and are used effectively. The example you used, chrome, uses update-alternatives for example (maybe not the rpm? the deb does).

I think this supports my (and Lennart's) point. Getting `update-alternatives` to work correctly is complicated. Two orders of magnitude more complicated than picking a non-conflicting name in a shared directory. (I had to deal with the fallout from sbin-merge intersecting with alternatives in Fedora and I don't want to ever return to this.)  So if the maintainers of that package can deal with that, they can certainly deal with putting the files of the package in the appropriate place under `/usr/lib`. They already _must_ deal with putting the files of the package in the appropriate laces under `/usr/bin`, `/usr/share/man`, `/usr/share/appdata`, `/usr/share/applications`, `/usr/share/icons` and whatnot. Prettending that the package is installed just under `/opt` and doesn't have to interact with the rest of the system is pointless.

specs/linux_file_system_hierarchy.md
```diff
@@ -34,6 +34,10 @@ Some directories like `/var/spool/` are not covered,
 even though it might make sense to include them
 in the structure of an actually deployed OS.
 
+Some directories are described for compatibility with current Linux distributions,
+but their use is not recommended.
+The subsections that describe those directories are marked with ⚠️.
```
brauner, line 39: I would add a separate heading for the section describing directories that are not recommended to make the separation stronger. Maybe you can send a follow-up for that?

PR #156, opened 2025-08-16
LFSH: add paragraph for /opt/

bluca:
/opt/ is used by software such as Chrome or Signal, and it's also supported by specifications like sysexts, so add a paragraph defining it.

### DIFF
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -80,6 +80,11 @@ and should fall back to defaults if configuration is missing,
 following the
 [Configuration Files Specification](configuration_files_specification.md).
 
+If `/opt/` is supported (see below),
+then the `/etc/opt/` subdirectory is the location where third-party software installed in `/opt/` may store
+their configuration.
+The same naming convention as directories under `/opt/` is used for directories under `/etc/opt/`.
+
 ### `/home/`
 
 The location for normal user's home directories.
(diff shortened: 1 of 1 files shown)


specs/linux_file_system_hierarchy.md
```diff
@@ -80,6 +80,11 @@ and should fall back to defaults if configuration is missing,
 following the
 [Configuration Files Specification](configuration_files_specification.md).
 
+If `/opt/` is supported (see below),
+then the `/etc/opt/` subdirectory is the location where third-party software installed in `opt` store their
```
behrmann, line 84: Is `/etc/opt/` a thing for anything but chrome? I was about to ask whether it was a thing at all, but then found a machine with a directory `/etc/opt/chrome`. I'm asking because I wonder whether it makes sense to spec, since software using `/opt/` will most likely not care for any other standards either.

specs/linux_file_system_hierarchy.md
```diff
@@ -80,6 +80,11 @@ and should fall back to defaults if configuration is missing,
...
+configuration.
+The same naming convention as directories under `/opt` is used for directories under `/etc/opt/`.
```
behrmann, line 86: More consistent slashes.
```suggestion
then the `/etc/opt/` subdirectory is the location where third-party software installed in `/opt/` store their
configuration.
The same naming convention as directories under `/opt/` is used for directories under `/etc/opt/`.
```

specs/linux_file_system_hierarchy.md
```diff
@@ -93,6 +98,16 @@ Applications should generally not reference this directory directly,
 but via the per-user `$HOME` environment variable,
 or via the home directory field of the user database.
 
+### `/opt/`
+
+The location for third-party vendors directories.
```
behrmann, line 103:
```suggestion
The location for third-party vendors' directories.
```
or
```suggestion
The location for third-party vendor directories.
```

specs/linux_file_system_hierarchy.md
```diff
@@ -93,6 +98,16 @@ Applications should generally not reference this directory directly,
...
+Third-party vendors (i.e.: unrelated to the OS provider) install a directory,
+typically named after the vendor or the software, under this location.
+Usually read-only, but this is not required.
```
behrmann, line 106: The first half should have a verb as well
```suggestion
It is usually read-only, but this is not required.
```

PR #163, opened 2025-09-09
BLS: say that initrd may be split at whitespace for compat

keszybz:
A path is not allowed to contain whitespace (only ASCII upper and lower case characters, digits, "+", "-", "_" and "." per the first section), so we know that a space cannot occur in the value. But unfortunately, this is an easy mistake to make. We made the situation worse by allowing other keys to contain whitespace-separated paths, e.g. 'devicetree-overlay' is defined as "multiple overlays are separated by spaces and applied in the same order as they are listed."  So it's easy to try to apply the same pattern in other places. Also, the implementation that was done for grub2 allowed multiple paths to be specified. It is certainly confusing if we say that
```
  initrd /some/path/initrd /some/path/extra
```
is not a valid. If instead we were to correctly understand that the user meant two different initrds and emit a warning, our behaviour would be more robust and user-friendly.

C.f. https://github.com/systemd/systemd/issues/38834
 https://bugzilla.redhat.com/show_bug.cgi?id=2323348
 https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/AYB4VHQ5K6CEV4NYCK2QTZB2OEAIJMP7/


specs/boot_loader_specification.md
```diff
@@ -264,13 +264,18 @@ The following keys are recognized:
...
-  The value is a path relative to the root of the file system
-  containing the boot entry snippet itself.
+* `initrd` specifies an initrd to use when executing the kernel (`cpio` image).
+  The value is a path
+  relative to the root of the file system containing the boot entry snippet itself.
   This key may appear more than once,
   in which case all specified images are used,
   in the order they are listed.
 
+  To increase compatibility,
```
behrmann, line 274: This begs the question "to increase compatibility with what?", since making this implementation-defined, might lead to more incompatibility between different implementations. Judging from the rationale in the commit message, it's to be more compatible with the semantics of other keys, but that's not clear form the text here alone.

I also think that this should note the preferred behaviour, so that implementations may converge on it.

PR #164, opened 2025-09-12
DDI: Add recomendation for LUKS encrypted partitions

frederictobiasc:
Add note clarifying LUKS encryption vs dm-verity protection to inform users not to confuse confidentiality with integrity and authenticity.

Addresses https://github.com/systemd/systemd/pull/38854#issuecomment-3281055110

### DIFF
--- a/specs/discoverable_disk_image.md
+++ b/specs/discoverable_disk_image.md
@@ -14,10 +14,17 @@ extensions, portable services, containers and more, and shall be protected by si
 into one.  They are designed to be composable and stackable, and provide security by default.
 
 ## Image Format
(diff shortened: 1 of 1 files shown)


poettering:
so thing about the combination of dm-crypt + dm-verity is that right now the most prominent implementations of DDIs, i.e. systemd's, doesn't support it. I am always reluctant to merge specs without code somewhere that implements it. After all, I am not entirely sure which of the two should be on top of the other, and my guess is thta we'd learn once we actually implement this. (my guess is dm-verity fist, dm-crypt second)

PR #175, opened 2025-10-27
specs/lfsh: describe /usr/lib and /usr/lib64 more fully

keszybz:
The spec so far described the multiarch scheme used in Debian and derivatives. This scheme is great, but for various reasons it hasn't seen wide adoption, and this means that the spec describes the layout that most users will actually see only very briefly. The spec also implied that the rules for multiarch and multilib systems are the same, just with sligtly different paths. But this is not entirely true, since on multilib systems /usr/lib plays a similar role as directories one level lower (i.e. /usr/lib/<arch-id>) on multiarch systems.

Since it's useful to describe the layouts that are commonly used, add a new section called "Libraries, system programs, and program assets" and describe `/usr/lib`, `/usr/lib/<arch-id>`, `/usr/lib64`, `/usr/libexec` there. Explicitly say that multiarch is the recommended approach, but others are commonly used too. I think this makes it easier for users to understand that this zoo of locations is a bikeshedding exercise to a large extent.

Also, change `/usr/lib/arch-id` to `/usr/lib/<arch-id>` everywhere.

The description of `/usr/share` is adjusted a bit. Previously, we said that this is is the location for data shared between programs and is subject to interoperability requirements, but quite obviously, a large part of /usr/lib plays the same role, with data that is shared between programs and subject to interoperability requirements. So copy the text about interoperability to the new section too, and say that /usr/share is for architecture-independent data. (This is the historical origin of "share" in the name.)

The general goal of this change is to make the spec more suitable for multilib systems.

### DIFF
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -214,7 +214,7 @@ Using a subdirectory is recommended if the package installs multiple files.
 It is also necessary if multiple versions of the same headers shall be coinstallable.
 The subdirectory may be named after the package or project providing it.
 Packages may place architecture-dependent header files and directories in a
-`/usr/include/arch-id/` subdirectory,
+`/usr/include/<arch-id>/` subdirectory,
 following the identifiers defined on the
 [Multiarch Architecture Specifiers (Tuples)](https://wiki.debian.org/Multiarch/Tuples)
 list.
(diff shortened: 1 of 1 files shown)


specs/linux_file_system_hierarchy.md
```diff
@@ -229,17 +229,65 @@ and other compiler-specific documentation.
...
-### `/usr/lib/`
+### Libraries, system programs, and program assets
+
+This section describes the directories used to store shared libraries,
+internal binaries or other programs that are not regularly invoked from a shell
+(and thus should not be placed in the directories in `$PATH`),
+and other static files that are part of program installations.
+
+There are two main schemes for shared libraries.
+On "multiarch" systems,
```
behrmann: Could we go for something else, e.g.
```suggestion
On *multiarch* systems,
```
I think that stands out a bit easier, doesn't have an additional glyph when grepping in the rendered text and would translate more nicely to other output formats.

Same below.

bluca (approved):

PR #181, opened 2025-11-14
Add note about canonical order of UKI components

arianvp:


specs/unified_kernel_image.md
```diff
@@ -47,6 +47,10 @@ header to 0x0A (i.e. `IMAGE_SUBSYSTEM_EFI_APPLICATION`).
 ## UKI Components
 UKIs consist of the following resources:
 
+<!--
+NOTE: these components are in canonical order for predictable PCR measurements.
+Please add any new components at the bottom of the list and NEVER reorder anything in this list.
```
bluca: Actually I think pcrsig and pcrpkey are the exception to this? IE, new things should be added before .pcrsig, but these two are never used for measurements

poettering:
Any chance you can add a brief comment to the visible part of the spec as suggested? seems good to go otherwise

keszybz:
I updated the patch to have both the comment and the visible text.

brauner (approved):

Issue #185, opened 2025-11-20
Suggestion: Numbering our specs


poettering:
I think we should start numbering our specs, we already have quite a few. It makes it easier to reference them, and gives a hint of a timeline between them.

my proposal: let's start naming them UAPI.1, UAPI.2, UAPI.3 and so on. For the existing ones let's retroactively assign numbers based on the order of the timestamp of the commits that added them. If multiple specs were added in a single commit, assign them numbers based on alphabetical order of the markdown file name.

Inspiration for this: the IETFC RFC process, or the oxide computing RFD process.

It has been suggested to pad the numbers to four digits with 0, to make sorting easier. Personally I am not convinced this is necessary, and it has the drawback that the name is harder to type, and I doubt people would always pad anyway.

PR #187, opened 2025-11-21
specs: assign numbers to all current UAPI specs

poettering:
Fixes: #185

### DIFF
--- a/specs/UAPI.1.md
+++ b/specs/UAPI.1.md
@@ -1,0 +1,1 @@
+boot_loader_specification.md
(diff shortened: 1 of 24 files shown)


bluca:
How does this look when rendered on the website? If we get duplicated entries on the left hand side pane it might not look too pretty: https://uapi-group.org/specifications/

bluca:
I think the list is auto-generated from the directory, so they'd probably get duplicated. What about adding it just to the title?

PR #202, opened 2025-12-18
specs: import OSC 3008 spec from systemd

poettering:
This stuff is really generic, and fits the purpose of the UAPI group, hence let's migrate this here.

This is mostly a 1:1 copy of the spec from systemd, but I enclosed the final part that carries a list of OSC sequences currently known in markdown comments, because it's not really part of the spec, but still kinda useful for spec writers.

(Maybe one day we should turn that table into a "registry" style spec of its own?)

### DIFF
--- a/specs/osc_context.md
+++ b/specs/osc_context.md
@@ -1,0 +1,346 @@
+---
+title: "UAPI.15 OSC 3008: Hierarchical Context Signalling"
+category: Concepts
+layout: default
+version: 1.0
+SPDX-License-Identifier: CC-BY-4.0
+weight: 15
+aliases:
+- /UAPI.15
+- /15
+---
+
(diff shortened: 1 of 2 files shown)


specs/osc_context.md
```diff
@@ -0,0 +1,346 @@
+---
+title: UAPI.15 OSC 3008: Hierarchical Context Signalling
```
t-lo:
```suggestion
title: "UAPI.15 OSC 3008: Hierarchical Context Signalling"
```
Not tested but this might fix it.

brauner (approved):

bluca (approved):

poettering:
Also see → https://github.com/systemd/systemd/pull/40132

PR #209, opened 2026-03-03
tpm-pcr-registry: document systemd extending PCR 7 with leave-initrd

src-up:
PCR7 is only extended on leave-initrd, so it differs in the main
OS. This enables distinguishing initrd vs OS when sealing keys to PCR7
(e.g. for systemd-repart-created volumes).

Since https://github.com/systemd/systemd/pull/40914


keszybz:
Replaced by https://github.com/uapi-group/specifications/pull/219.

PR #212, opened 2026-04-21
UAPI.1: introduce "extra" stanza

poettering:
This corresponds to the new logic implemented in
https://github.com/systemd/systemd/pull/41543: it allows type1 entries to pass parameterization via sysexts/confexts/creds similar to how we support this for type2 already.

### DIFF
--- a/specs/boot_loader_specification.md
+++ b/specs/boot_loader_specification.md
@@ -361,6 +361,16 @@ The following keys are recognized:
 
   Example: `profile 1`
 
+* `extra` refers to additional resources to pass to the invoked kernel. This takes a path relative to the
+  root of the file system containing the snippet itself, referring to a regular file. The file name must be
+  suffixed with a recognized suffix which indicates the type of additional resource. Currently, recognized
+  suffixes are `.confext.raw` (for systemd-style configuration extension DDIs), `.sysext.raw` (for
+  systemd-style system extenion configuration extension DDIs) and `.cred` (for systemd-style encrypted system
+  credential files). In future additional suffixes will be defined and implementations must gracefully handle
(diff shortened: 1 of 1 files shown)


specs/boot_loader_specification.md
```diff
@@ -373,6 +383,8 @@ key. Here is an example for a complete drop-in file:
     architecture x64
     linux        /6a9857a393724b7a981ebb5b8495b9ea/3.8.0-2.fc19.x86_64/linux
     initrd       /6a9857a393724b7a981ebb5b8495b9ea/3.8.0-2.fc19.x86_64/initrd
+    extra        /6a9857a393724b7a981ebb5b8495b9ea/3.8.0-2.fc19.x86_64/somedata.cred
+    extra        /6a9857a393724b7a981ebb5b8495b9ea/3.8.0-2.fc19.x86_64/somethingelse.cred
```
brauner, line 387: I'm a bit lukewarm on the "extra" terminology. Maybe "add" or "addon"?

poettering: we have four kinds of "extras":

1. addons (which are SB signed EFI PE binaries with additional initrds, device trees…)
2. confexts (which are DDIs)
3. sysext (ditto)
4. credentials (which are their own thing)

Hence, the word "addon" is already used for something very specific in the UKI context, and the word "extra" is used as a way of grouping all of these sidecar concepts under a common term.

Also note that we already support that each UKI `foobar.efi` can have a sidecar dir `foobar.efi.extra.d/` where you can add such extras, and we used the name there already.

or in other words: this just reuses the existing concepts under their existing names and makes them available in type 1 boot loader spec snippets too, where they previously have been available only in type 2 UKIs.

Issue #207, opened 2026-02-11
Please develop a filesystem specification (fsspec)


bdrung:
There are multiple tools (like cpio, 3cpio, tar, mksquashfs, etc.) that create a filesystem archive. Many of those tools can take a directory or tarball is input. User (like dracut and initramfs-tools) create the content of the archive in a temporary directory and then call the archive tool. This step could be avoided if the archive tools would support a way to specify the content and map filenames. Sadly every tool supports a different way to (partially) do that:

* `3cpio` supports a manifest file: https://github.com/bdrung/3cpio/blob/main/man/3cpio.1.adoc#manifest
* `gensquashfs` (from squashfs-tools-ng) support a pack file: https://manpages.debian.org/trixie/squashfs-tools-ng/gensquashfs.1.en.html#PACK_FILE_FORMAT
* `mkfs` has a special proto format: https://www.unix.com/man-page/v7/1M/mkfs
* `mkproto` has prototype files: https://www.ibm.com/docs/en/aix/7.3.0?topic=m-mkproto-command
* `mksquashfs` (from squashfs-tools >= 4.5) support a pseudo file: https://manpages.debian.org/unstable/squashfs-tools/mksquashfs.1.en.html#PSEUDO_FILE_DEFINITION_FORMAT
* `tar` has `--transform`
* The Linux kernel has `gen_init_cpio`: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/usr/gen_init_cpio.c#n570

Can we agree on a specification that would satisfy all those tools and use UAPI for publishing such spec? Some discussion has started in https://github.com/erofs/erofs-utils/issues/37 already. I am the author of 3cpio and would love to switch to a common format.

From looking at the existing formats, I propose following properties:

* text file (probably UTF-8)
* line based (one line per file) (no JSON or such)
* allows comments (e.g. starting with #)
* bonus point: extensible
* bonus point: constructable with `find . | sed '$simplerule'`

### One possible format

```
#fsspec v1
# First line is a filesystem specification marker (needed for detection and highlighting)
# comments start with #
<filename> f [mode:<mode>] [uid:<uid>] [gid:<gid>] [mtime:<mtime>] [size:<filesize>] [source:<souce_filename>]
<filename> d [mode:<mode>] [uid:<uid>] [gid:<gid>] [mtime:<mtime>]
[...]

PR #213, opened 2026-04-22
Add UAPI.16: File Manifest spec (WIP)

poettering:


poettering:
(mostly posted here to start the discussion)

bluca:
As mentioned elsewhere, it would be great if this could be embedded in existing json manifests, to avoid having to ship multiple ones, and consumers knew how to find it - essentially the existing mkosi manifest. If I understand correctly, the only thing needed for this to work is to establish an optional and well-known "key" under which this object can be found under a parent json object?

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+| Version | Changes         |
+|---------|-----------------|
+| 1.0     | Initial Release |
+
+This format stores information about static, immutable data resources that can be acquired and stored on a
+POSIX file system or written to a GPT disk partition.
+
+This manifest file format is inspired by the traditional UNIX `SHA256SUMS` and BSD `mtree(5)` file formats,
+but supersede them in various ways:
```
behrmann:
```suggestion
but supersedes them in various ways:
```

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+1. It allows declaring encoding of data files (i.e. `gzip` compression and suchlike)
```
behrmann:
```suggestion
1. It allows declaring the encoding of data files (i.e. `gzip` compression and suchlike)
```

behrmann: Also all of these points are full sentences and already start capitalised, so a period at the end would make sense.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+```json
+{
+        "mediaType" : "application/vnd.uapi.manifest"
+        "files" : [ … ]
+}
+```
+
+### Notes
+
+The `files` field is just an array of file objects, see below.
```
behrmann: Style not: Drop the just, it's only baggage for what you want to say
```suggestion
The `files` field is an array of file objects, see below.
```

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+```json
+{
+        "name" : "…",
+        "dataEncoding" : "gzip"
+        "encodedDataSize" : 2011,
+        "dataSize" : 4711,
+        "dataFile" : "…",
+        "dataUrl" : "http://…"
+        "dataLiteral" : "base64…"
```
behrmann: I think dropping the `base64` here makes this a bit clearer just form looking at it, since my first reaction upon seeing this was to check below whether this is a necessary prefix
```suggestion
        "dataLiteral" : "…"
```

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+        "gptFlagGrowFileSystem" : true,
+        "readOnly" : true,
+        "validFromUSec": 4711,
+        "validUntilUSec" : 4711,
+}
+```
+
+### Notes
+
+All fields but `name` are optional.
```
behrmann: Does optional here mean that they can be omitted, they can be `null` or both?

Generally, it would be nice to write the expected types values either here or in the JSON above, so one doesn't have to guess it from the example, since `validUntilUSec` allows the string `"infinity"`.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+## Manifest Object
+
+This is the top-level object for a manifest file. It's designed to be
+extensible and self-identifies as manifest.
+
+```json
+{
+        "mediaType" : "application/vnd.uapi.manifest"
+        "files" : [ … ]
+}
```
pothos:
```suggestion
        "forced": ["name", …]
}
```

*Notes* text:

The optional `forced` array holds the `name` values a client should look up in `files` and use even if it's on a higher version.

Reason:

We currently don't have a way to communicate a downgrade when we notice that an update is faulty.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+        "validUntilUSec" : 4711,
+}
+```
+
+### Notes
+
+All fields but `name` are optional.
+
+The `name` must be a valid UTF-8 POSIX filename, may not contain control characters (ASCI 0…31, 127), may not
+contain a slash, and may not be identical to `.`, `..`, ``. It may have a maximum length of 255 bytes.
```
behrmann: > may not be identical to `.`, `..`

Just to be on the safe side: So it's fine for them to start with `.` and `..`, e.g. `.foo` and `..bar` are okay?

> It may have a maximum length of 255 bytes.

Does that mean it is defined by the implementation what to do if the name is longer than 255 bytes?

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+Only one of `dataFile`, `dataUrl`, `dataLiteral` should be set. dataFile references a file in the same place
+as the manifest, dataUrl a full URL, and dataLiteral inline Base64.
+
+`sliceOffset`, `sliceSize` are apply to the *decoded* data blob. If not specified explicitly should default
```
behrmann:
```suggestion
`sliceOffset`, `sliceSize` are applied to the *decoded* data blob. If not specified explicitly should default
```

specs/file_manifest.md
```diff
@@ -0,0 +1,219 @@
...
+        "sliceSize" : 33,
+        "sha256" : "…",
+        "gptLabel" : "…",
+        "gptTypeUuid" : "…",
+        "gptFlagNoAuto" : true,
+        "gptFlagGrowFileSystem" : true,
+        "readOnly" : true,
+        "validFromUSec": 4711,
+        "validUntilUSec" : 4711,
+}
```
pothos, line 112:
```suggestion
        "steppingStone" : false,
}
```

_Notes_ text:

`steppingStone` may be set to indicate that the client should not just take the highest version it can but instead it should first update to the lowest version marked as "stepping stone" but higher than the client's current version.

Reason:

Sometimes a breaking change is required which means than an old client can't consume the version with the breaking change. Instead, it first needs to update to a version that ensures compatibility with the update payload. For very old clients this means they work their way through a couple of updates to finally reach the lastest version.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+`dataEncoding` should use the same encoding specifiers as HTTP.
+
+`sha256Sum` is the SHA256 hash of the specified slice, formatted in 64 hexadecimal characters. Parsers should
+parse this case-insensitively.
+
+The `gpt*` fields encode fields that we need when placing these resources in a GPT partition table.
+
+If `gptLabel` is unspecified, and the data is written to a GPT partition it should use the value of `name` as
+label.
```
behrmann: If the label is what's referred to as [`PartitionName`](https://uefi.org/specs/UEFI/2.10/05_GUID_Partition_Table_Format.html) here the maximum length is 72 bytes. So this should maybe say that it needs to be truncated in the default case if linger.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+`readOnly` is relevant both for GPT partition table and should be reflected in the 'w' access bit when
+written to a regular file.
+
+`valid*USec` is in µs since UNIX epoch, defines a validity time window of the resource. The data should not
+be accessed or consumed outside of the specified time window. If `validFromUSec` is not specified it should
+default to 0, if `validToUSec` is not specified it should default to infinity.
```
behrmann:
```suggestion
default to 0, if `validUntilUSec` is not specified it should default to infinity.
```

Nit: Personally I think `validBeforeUSec` and `validAfterUSec` would be clearer.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+1. If `validFromUSec` is set and greater than the current time, the download should immediately fail.
+2. If `validUntilUSec` is set and lower than the current time, the download should immediately fail.
+3. If `dataLiteral` is set, it should be Base64 decoded, continue in step 7
+4. Otherwise, if `dataUrl` is set, the data from the URL should be acquired, continue in step 7
+5. Otherwise, if `dataFile` is set, the data should be acquired from the same URL as the manifest itself, however with the last component of the URL replaced by the file name encoded in `dataFile`, following the same semantics as HTML relative links. Continue in step 7.
+6. Otherwise, the data should be acquired as in step 3, but the `name` string should be used as file name to suffix the URL with.
+7. If `encodedDataSize` or `dataSize` are set: the size of the downloaded data shall be checked against `encodedDataSize` (if set) or `dataSize` (otherwise).
+8. If `encoding` is set: the downloaded data shall be decoded according to the algorithm indicated in `encoding`.
+9. if `dataSize` is set: the resulting decoded data shall be checked against `dataSize`.
```
behrmann:
```suggestion
9. If `dataSize` is set: the resulting decoded data shall be checked against `dataSize`.
```

Foxboron:
Is the intention of this specc to solve this issue, or is this trying to solve a different problem?

https://github.com/uapi-group/specifications/issues/207

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+        "mediaType" : "application/vnd.uapi.manifest"
+        "files" : [ … ]
+}
+```
+
+### Notes
+
+The `files` field is just an array of file objects, see below.
+
+We might eventually add additional fields to the manifest, for in-line signatures and similar.
```
behrmann: I tend to keep harking on this, but a version field would be nice so one could simply check "Oh, the version is 3, I can only handle version 1 and 2, guess I'll give up", but this could probably also be implemented as new media type `application/vnd.uapi.manifest2`, but then adding new fields wouldn't be necessary.

daandemeyer:
@Foxboron Totally different problem, it's designed to replace SHA256SUMS for sysupdate to list remote resources

Foxboron:
@daandemeyer Hmm, should the title be `Sysupdate File Manifest spec` to nail the usage a bit more down? Else we might end up with multiple "file manifest" specs?

specs/file-manifest.md
pothos:
```suggestion
        "tags": ["myext", …]
}
```

_Notes_ text:

The optional `tags` array holds strings that appear as part of the `name` values and should be used to match additional files in a dynamic way. E.g., sysupdate should use this to expand the `%t` specifier in a `MatchPattern=` of a `.transfer` file which then instantiates multiple file transfers, one for each `tags` entry. Cleanup should be handled by marking the written files with xattrs (the `MatchPattern` value itself or some other value set in the transfer file) so that sysupdate can discard those files in a directory that don't match anymore.

Reason:

Currently we can't add new configuration/extensions but have to first add new sysupdate transfer files to a client. As a hack one could today have a "singleton" confext that would ship the transfer files but that means one always has to do a 2 step update but there is no guarantee that then the update is done and one might to catch up again and again to avoid being in a partly updated state. The dynamic extension proposed here fixes this.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+---
+
+# UAPI.16 File Manifest
+
+| Version | Changes         |
+|---------|-----------------|
+| 1.0     | Initial Release |
+
+This format stores information about static, immutable data resources that can be acquired and stored on a
+POSIX file system or written to a GPT disk partition.
```
keszybz: I get the intent here, but the second part of the sentence is just so random. They can be written to a non-POSIXX file system, to a db, or stored in a zip file, or even written to DOS partition.

specs/file-manifest.md
keszybz: It also doesn't supersede them. Maybe "extends"?

specs/file-manifest.md
keszybz: "inline"

specs/file-manifest.md
keszybz: Please put quotes ("") around the `.`, `..`, and ``. (When rendered, `` is just an empty space. With quotes, it'll render much better.)

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+### Notes
+
+All fields but `name` are optional.
+
+The `name` must be a valid UTF-8 POSIX filename, may not contain control characters (ASCI 0…31, 127), may not
+contain a slash, and may not be identical to `.`, `..`, ``. It may have a maximum length of 255 bytes.
+
+Only one of `dataFile`, `dataUrl`, `dataLiteral` should be set. dataFile references a file in the same place
```
keszybz: `dataFile` and others should be marked up.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+`sha256Sum` is the SHA256 hash of the specified slice, formatted in 64 hexadecimal characters. Parsers should
+parse this case-insensitively.
+
+The `gpt*` fields encode fields that we need when placing these resources in a GPT partition table.
+
+If `gptLabel` is unspecified, and the data is written to a GPT partition it should use the value of `name` as
+label.
+
+`readOnly` is relevant both for GPT partition table and should be reflected in the 'w' access bit when
```
keszybz: "both" seems misplaced.

specs/file-manifest.md
```diff
@@ -0,0 +1,219 @@
...
+
+If neither `dataFile`, nor `dataUrl`, nor `dataLiteral` are specified, the `name` field should be used as
+fallback equivalent to it being repeated in `dataFile`.
+
+`dataLiteral` may be encoded either in regular Base64 or in URL-safe Base64, consumers must be able to deal
+with either.
+
+Additional hash algorithms may be defined in future.
+
+Sizes, offsets, timestamps shall use positive integer JSON numbers, and may use values above 2^52. (This
```
keszybz: "2²⁵" ?

keszybz (changes requested):
I think it'd be nice to use [semantic line breaks](https://sembr.org/) here. We agreed in general to do this in new documents.

specs/file-manifest.md
cyphar: In general I think priorities are a better way of doing this than having a binary "force" option -- this reminds me a little of the classic `!important` mistake in CSS (once you set something as `!important` you can't set anything as higher priority).

(In our setup we will also potentially have several layers of repos which map much better to priorities than a binary "force" option IMHO. Though we will probably add something like [CELs](https://cel.dev/) to filter out entries more dynamically...)

cyphar: If you're open to doing it this way, I'm all for it. We could eventually add some more of the `machine-info` stuff here, so maybe it should get a dedicated `overrides` object that then has keys matching the `/etc/machine-info` names?

My only concern is that while we will use this, I don't know if other users of this manifest spec would want this feature?

specs/file-manifest.md
poettering: I think for breaking changes we really should use a new media type. And of course, we should really avoid breaking changes in the first place.

I'll add a brief paragrpah about this.

poettering:
> @daandemeyer Hmm, should the title be `Sysupdate File Manifest spec` to nail the usage a bit more down? Else we might end up with multiple "file manifest" specs?

I'd probably call the other spec an "Inode Spec", since that's more what it is. Or "File System Object" or so.

This one here is purely about files.

You are absolutely right, this might be confusing, but it will either way I am sure.

I am against naming this after the software that likely implements this, it's supposed to be a generic spec, independent of any specific implementation.

poettering:
Hmm, we could also consider extending this spec to just cover what is requested in https://github.com/uapi-group/specifications/issues/207 too. I mean, if I grok this right it would just mean adding some more fields to encode UNIX inode properties in full: i.e. `inodeType`, `mode`, `uid`, `gid`, `user`, `group`, `major`, `minor`, `symlinkTarget` and so on. When using this for the download scenario we'd use ignore all these fields I guess, but I see nothing speaking against supporting this too.

specs/file-manifest.md
pothos: I thought that a simple solution here is good because it allows exact control of the version. This is meant for machines you control or to retract a broken update. Setting this would happen on the last layer when one, e.g., imports this manifest from somewhere and then prepares it for one's own infra. The upstream repo could also have this to retract an update but then on import one would be able to change this to even an older version.

Priorities sound more complex and opaque. Normally we want to be on the latest version, so what do priorities help with?

pothos: This should easily be usable for configuration management through confexts/sysext in a gitops way.

poettering: Hmm, why not just set a `revoked` boolean on the individual file instead: if set to true we'll downgrade to the next older version if nothing newer is available, but something older is.

poettering: I added `revoked` now, under the assumption this should work too.

poettering:
Posted a new version covering all comments, but not trying to address #207 (would prefer if we did that in a later follow-up PR)

bluca:
> As mentioned elsewhere, it would be great if this could be embedded in existing json manifests, to avoid having to ship multiple ones, and consumers knew how to find it - essentially the existing mkosi manifest. If I understand correctly, the only thing needed for this to work is to establish an optional and well-known "key" under which this object can be found under a parent json object?

^^^ ?

poettering:
> As mentioned elsewhere, it would be great if this could be embedded in existing json manifests, to avoid having to ship multiple ones, and consumers knew how to find it - essentially the existing mkosi manifest. If I understand correctly, the only thing needed for this to work is to establish an optional and well-known "key" under which this object can be found under a parent json object?

i don't grok this request?

it seems to me that the manifest format here could easily be embedded by mkosi's package manifests if it wants file-level information. But that's something to decide and define in mkosi's format, it's not something we could dictate here?

bluca:
> > As mentioned elsewhere, it would be great if this could be embedded in existing json manifests, to avoid having to ship multiple ones, and consumers knew how to find it - essentially the existing mkosi manifest. If I understand correctly, the only thing needed for this to work is to establish an optional and well-known "key" under which this object can be found under a parent json object?
> 
> i don't grok this request?
> 
> it seems to me that the manifest format here could easily be embedded by mkosi's package manifests if it wants file-level information. But that's something to decide and define in mkosi's format, it's not something we could dictate here?

Having a "suggested" key for the object would allow consumers to know what to search for, without having to come up with one for each case. Just like there's a "suggested" filename for the file in the spec?

specs/file-manifest.md
pothos: Yeah, for most cases this should be okay but I see a few where it's not the best solution.

- For a larger jump back one would have to add it to all newer versions but I guess since one anyway does this by postprocessing it's just a bit more compute, so not too bad. Can be annoying to script though.
- One would have to have an entry for the version the client is on which might not be the case when it's a custom build. This requires the update server to be smart and be able to have the client version reported to inject such an entry marked as revoked.

pothos: @poettering I meant the tag to be a global thing, not for each file. It should be used to decide on which files to look at.

specs/file_manifest.md
pothos: Thanks for the rewording of the text, looks good now!

specs/file-manifest.md
pothos: But I also see merit in `revoked` because it fit very well for the publisher use case. Since they focus on different things, can we have both the global `forced` array and the `revoked` boolean? Then one can pick the right one for the use case instead of having to do workarounds.

pothos:
I wouldn't merge this into the mkosi manifest but treat it like the SHA256SUMS file mkosi can generate with `--checksum=` because this update manifest is supposed to have a specific file name that sysupdate can look for while the mkosi manifest includes the version in its name and thus can't be used as is. So `--uapi-file-manifest=` would write it and then `--sign=` would do the gpg signing for this file like it's currently done for SHA256SUMS.

bluca:
Nah it has to be in the same file, at least as an option, as I most definitely do not want to have to deal with having to publish yet another file

specs/file-manifest.md
cyphar: Also, this information should live inside the files array, not be another top-level array imho.

AdrianVovk: I don't really like this idea, it's simply too magical and it's also kinda fragile. What if you want a transfer definition that's different than the preexisting template?

What is already possible today is that you have a sysupdate component that ships a sysext with additional .transfer files. If you decide to add a new confext/sysext available for download, you can add all the transfer definitions you need into that sysext and update it independently of the host system.

The patches that introduce update streams to sysupdate also add the ability to drop additional streams into /var/cache and it'll just load transfer files from there. By making it not-a-sysext and just teaching sysupdate to look for additional transfer files in a special directory, you can have a sysupdate component that downloads these additional transfers and then sysupdate can load them immediately without refreshing the sysexts

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+}
+```
+
+### Notes
+
+The `files` field is an array of file objects, see below.
+
+Files may appear in any order in the manifest, but it's recommended to sort them alphabetically by their name.
+
+The `name` fields of file objects are supposed to be uniquely assigned within each manifest file.
```
AdrianVovk: sounds like a JSON key in an object?

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+
+`tags` is an array of arbitrary usecase-specific strings. This may be used for filtering, for example to
+categorize files in various ways (for example, files could be tagged "unstable", "stable", "lto", to indicate
+stability levels, or similar).
+
+`revoked` is a boolean. If true, the file should not be requested anymore, and if it has been acquired
+before, appropriate steps should be taken to invalidate/remove it from use, if possible (details are
+implementation-specific). If unspecified, defaults to `false`.
+
+`steppingStone` is a boolean, defaulting to `false`. When this manifest file is processed by a software
```
AdrianVovk: Let's say v2 is a steppingStone and v3 is available, and a client system is on v1.

In practice, how would this be implemented? I assume what would have to happen is that sysupdate simply doesn't list `v3` as an available update. Effectively `v2` is the latest available update for any system running `v1`

So perhaps we should clarify the behavior here, that the updater will not advertise any versions newer than the first steppingStone it encounters that's applicable to the given client system

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+`validAfterUSec` is not specified it should default to 0, if `validBeforeUSec` is not specified it should
+default to the largest supported unsigned integer value.
+
+If neither `dataFile`, nor `dataUrl`, nor `dataLiteral` are specified, the `name` field should be used as
+fallback equivalent to it being repeated in `dataFile`.
+
+`dataLiteral` may be encoded either in regular Base64 or in URL-safe Base64, consumers must be able to deal
+with either.
+
+`tags` is an array of arbitrary usecase-specific strings. This may be used for filtering, for example to
```
AdrianVovk: I'm not totally understanding how this is supposed to work. Things like "unstable"/"stable"/etc is best handled by update streams anyway

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+The `name` must be a valid UTF-8 POSIX filename, may not contain control characters (ASCII 0…31, 127), may
+not contain a slash, and may not be identical to "`.`", "`..`", "``" (but may contain these strings). It may
+have a maximum length of 255 bytes. Files whose names do not match these rules cannot be encoded in manifest
+files defined by this specification.
+
+Only one of `dataFile`, `dataUrl`, `dataLiteral` should be set. `dataFile` references a file in the same
+place as the manifest file itself, `dataUrl` a full URL, and `dataLiteral` inline Base64. `dataUrl` may use
+`http://…` and `https://…` schemes only.
+
+`dataSize` is the size of the decoded blob, `encodedDataSize` of the encoded blob. If `dataEncoding` is not
```
AdrianVovk: How does `dataSize` interact with `sliceSize` if both are set?

Perhaps we should just define a `dataOffset` and use that along with `dataSize` instead of having separate `slice*` fields?

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+
+The `gptLabel` field may not be longer than 72 characters. If `gptLabel` is unspecified, and the data is
+written to a GPT partition it should use the value of `name` as label instead. Note that this might require
+mangling, as GPT partition labels have smaller size limits than file names.
+
+`readOnly` (a boolean) is relevant for initializing the read-only bit of GPT partition table entries. It
+should also be reflected in the POSIX 'w' access mode bit when storing the data in a regular file on disk. If
+unspecified, defaults to false.
+
+`valid*USec` is in µs since UNIX epoch UTC, and shall be an unsigned integer. It defines a validity time
```
AdrianVovk: We can expire individual files this way, but what we really aught to be able to do is expire the entire manifest

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+implemented.
+
+1. If `revoked` is set to true, the download should immediately fail.
+2. If `validAfterUSec` is set and greater than the current time, the download should immediately fail.
+3. If `validBeforeUSec` is set and lower than the current time, the download should immediately fail.
+4. If `dataLiteral` is set, it should be Base64 decoded, continue in step 8.
+5. Otherwise, if `dataUrl` is set, the data from the URL should be acquired, continue in step 8.
+6. Otherwise, if `dataFile` is set, the data should be acquired from the same URL as the manifest itself, however with the last component of the URL replaced by the file name encoded in `dataFile`, following the same semantics as HTML relative links. Continue in step 8.
+7. Otherwise, the data should be acquired as in step 5, but the `name` string should be used as file name to suffix the URL with.
+8. If `encodedDataSize` or `dataSize` are set: the size of the downloaded data shall be checked against `encodedDataSize` (if set) or `dataSize` (otherwise).
```
AdrianVovk: If there's a `dataEncoding` but no `encodedDataSize`, you can't check the downloaded size against `dataSize`.

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+`xMyProjectWeight`.
+
+## Future Revisions
+
+This specification is intended to be incrementally improved. Additional fields may be defined at any time,
+with additional, optional metadata. Should a breaking change be necessary the media type will be changed.
+
+A number of future extensions are envisioned:
+
+* Inline cryptographic signatures
```
AdrianVovk: We should have this in the manifest from day 1. Part of what we want to do here is get away from GPG, right?

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+written to a GPT partition it should use the value of `name` as label instead. Note that this might require
+mangling, as GPT partition labels have smaller size limits than file names.
+
+`readOnly` (a boolean) is relevant for initializing the read-only bit of GPT partition table entries. It
+should also be reflected in the POSIX 'w' access mode bit when storing the data in a regular file on disk. If
+unspecified, defaults to false.
+
+`valid*USec` is in µs since UNIX epoch UTC, and shall be an unsigned integer. It defines a validity time
+window of the resource. The data should not be accessed or consumed outside of the specified time window. If
+`validAfterUSec` is not specified it should default to 0, if `validBeforeUSec` is not specified it should
```
AdrianVovk: We should explicitly specify that if there's no `validBeforeUSec` then the file just never expires. Rather than obliquely talking about "largest supported unsigned integer value"; that's not super clear

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+        "readOnly" : true,
+        "validAfterUSec": 4711,
+        "validBeforeUSec" : 4711,
+        "tags" : ["tag1", "tag2", …],
+        "revoked" : false,
+        "steppingStone": false,
+}
+```
+
+### Notes
```
AdrianVovk: Should we have a `version` field? Since we pulled out so many of the other `@` match patterns out into fields here. Or I guess since `@v` is a required pattern we want to keep encoding it into the filename?

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+8. It's an extensible format permitting vendors and projects to add their own per-file and per-manifest fields.
+
+This format is designed with tools such as `mkosi` and `systemd-sysupdate` in mind, but it's intended to be
+generally useful as a way to describe collections of files that may be acquired over the network or from
+local storage, and verified cryptographically.
+
+## Manifest File Name and Media Type
+
+When stored in a file system directory – alongside the data files it references – the manifest file should be
+named `Uapi16ManifestFile`.
```
AdrianVovk: I'd prefer we drop the file off the end of the name.

For one, naming a file "SomethingFile" is kinda redundant

For two, the HTTP endpoint might not actually be a file. If we want to put something smart behind the HTTP endpoint then the data can come from anywhere!

specs/file-manifest.md
AdrianVovk: > One would have to have an entry for the version the client is on which might not be the case when it's a custom build

Why would that be needed? If your on a custom build, then your version number is still either > or < than the version numbers in the manifest. Your exact version number doesn't need to appear in there

specs/file-manifest.md
AdrianVovk: Though on the other hand, I do actually quite like the fact that the file object is self contained. This means that when we factor out the download backends from sysupdate we can pass in this whole object

AdrianVovk:
> Nah it has to be in the same file, at least as an option, as I most definitely do not want to have to deal with having to publish yet another file

Sorry I'm not groking how you can avoid this. `mkosi` spits out a manifest listing the specific packages it included in a specific image that it built. This spec is about listing all the files in the directory, and in the sysupdate usecase that means across version boundaries. So if anything, the mkosi manifest would have to be embedded in here (describing mkosi's output files) rather than the other way around. But that also doesn't make sense because even if mkosi spits out multiple files you'll get just one manifest, right?

Anyway, this isn't "yet another file" to publish since it'll replace SHA256SUMS

specs/file-manifest.md
lnussel: Ack. I'm concerned that inline signatures expose a json parser to untrusted, potentially malicious data. To verify signatures, the signatures themselves need to be skipped so this might require some data mangling. Not sure that can be done safe enough in this context.

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+        "mediaType" : "application/vnd.uapi.16.file.manifest",
+        "files" : [ … ]
+}
+```
+
+### Notes
+
+The `files` field is an array of file objects, see below.
+
+Files may appear in any order in the manifest, but it's recommended to sort them alphabetically by their name.
```
brauner: We either should manadate a canonical order or we should just say "any order". Because the way it is phrased right now what will end up happening is that most tools will sort alphabetically and then you have 3 cases that don't and hilarity ensues.

specs/file_manifest.md
```diff
@@ -0,0 +1,270 @@
...
+            "sliceOffset" : 351272960,
+            "sliceSize" : 234003200,
+            "sha256" : "00b70a0813e15f309828d3a36156283cba87576c26755e0b2d4cf0951eff8163",
+            "gptLabel" : "FooOS_root",
+            "gptTypeUuid": "4f68bce3-e8cd-4db1-96e7-fbcaf984b709",
+            "readOnly": true,
+            "validBeforeUSec" : 1776856773123234
+        }
+    ]
+}
```
brauner, line 407: Let's say someone wants to use this file to describe an ML dataset that has billions of files. Will that scale?

brauner: cuI guess because you can have links to other files it wouldn't be a problem.
Curious though: can a manifest point to another manifest?

specs/file-manifest.md
pothos: If it's a dev build with a 9999 version this this won't do a downgrade when the offered versions are lower. One would have to add a 9999 version entry that is revoked to trigger the downgrade.

pothos: As for top-level or part of a file, I think it's easier to ensure that there is only one "forced" setting if it's a global rather than an attribute for a file. Or did you mean that "files" should not be an array but a dict so that it could go there but still be global?

pothos: First having to ship transfer files as update and then having to do one more update is the problem to be solved here. One can't automate this well to know that one is done.

To make it even more explicit when a transfer file is a template we can name it `NAME@.transfer` similar to how it's done for instantiating units. Only in these it would be allowed to use `%t`. But I do realize now that this assumes that all the parts pulled in this way would have to advance their version together when they are all in one component which for sysexts means that the actual software version is not visible anymore. So I guess it would be more flexible to say that components can be instantiated from `/etc/sysupdate.NAME@.d/some.conf` for each tag (The transfer file would then be allowed to use `%t` and the same xattr mechanism can do the cleanup). This allows the versions to be independent.

pothos: > What if you want a transfer definition that's different than the preexisting template?

Right, the mechanism assumes that there are supported categories, e.g., confext and sysext images or portable services. But anything new would need an update of the base OS (or whatever manages the transfer files), e.g., like suddenly you want VM images, too. I think that's acceptable because for every new kind of update one also has to think about how this gets applied. E.g., a new portable service image can't just be dropped, and one also has to define whether it should be loaded live or through a reboot.

specs/file-manifest.md
pothos: Yes, the simplest implementation could do it like that. I see it as a UX detail - ideally sysupdate would be able to display that there is an even newer version but it's blocked ("chained"!?).

specs/file-manifest.md
pothos: Yes, it's not for streams but to be able to add new files to a machine without having to first ship it new .transfer files. The high level story is that I want to manage a machine and first it starts with a confext/sysext for k8s and one for wireguard but now I want to add a monitoring confext and maybe drop wireguard. As said in the other comment thread, managing these transfer files through a confext itself is a rather ugly and fragile hack. Not only because of the 2 step update, or maybe 3 or even more, because one could always land "in between", but also because this gives you no way of cleaning things up again that were added. Hence the idea for a first class concept that allows to have a reliable update mechanism and a way for cleanup, too (through xattr marking).

Maybe best is if this spec would go together with a sysupdate implementation. But just having to tweak words for now than code also has it's benefits.

specs/file-manifest.md
pothos: Detached signatures aren't too bad though, or? There are alternatives to GPG.

specs/file-manifest.md
pothos: For this my idea was to have it as global property. Having it for a file won't work for the use case. I think one could still have tags but I agree I don't see how this here would help because one could already put this info into the filename itself so that sysupdate would match for it.

specs/file-manifest.md
pothos: Oh, the PR for streams uses `sysupdate@*.d` which looks similar to `sysupdate.*@.d` and one is for the host/system and the other is for a component.

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+`dataEncoding` should use the same encoding specifiers as HTTP.
+
+`sliceOffset`, `sliceSize` (unsigned integers) are applied to the *decoded* data blob (i.e. after
+decompression). If not specified explicitly, they should default to a zero offset, and a slice size equal to
+the full decoded size.
+
+`sha256` is the SHA256 hash of the specified slice, formatted in 64 hexadecimal characters. Parsers should
+parse this case-insensitively.
+
+The `gpt*` fields encode fields that we need when placing these resources in a GPT partition table entry. The
```
aplanas: I am not sure that the `gpt*` fields belong to the `file` object.  The gpt data of one object can contradict another one and there are not clear rules about how to resolve them.

Maybe is part of the manifest root?  Maybe part of a different UAPI 16 Container Manifest spec?

bluca:
> > Nah it has to be in the same file, at least as an option, as I most definitely do not want to have to deal with having to publish yet another file
> 
> Sorry I'm not groking how you can avoid this. `mkosi` spits out a manifest listing the specific packages it included in a specific image that it built. This spec is about listing all the files in the directory, and in the sysupdate usecase that means across version boundaries. So if anything, the mkosi manifest would have to be embedded in here (describing mkosi's output files) rather than the other way around. But that also doesn't make sense because even if mkosi spits out multiple files you'll get just one manifest, right?

The "top-level" build is mkosi, and it's a single mkos build that produces these artifacts. Just like it right now includes extensions DDI metadata in the mkosi manifest, it can also include this metadata.
The other way around would break existing consumers, which would suddenly not find the objects they are looking for anymore. New consumers looking for the new manifest would have no problem though, as they are new.

> Anyway, this isn't "yet another file" to publish since it'll replace SHA256SUMS

Well it is, as it's new

specs/file-manifest.md
poettering: hmm? not sure i follow? how would they contradict? each other? these are gpt per-partition fields?

poettering:
> Having a "suggested" key for the object would allow consumers to know what to search for, without having to come up with one for each case. Just like there's a "suggested" filename for the file in the spec?

I am not convinced that is necessary. For example, let's say we do something to address #207. The way I see this happen would be that for directories they'd contain an additional `contents` field in the file object that itself is again a manifest object. i.e. you'd nest them nicely. I think it would be really weird to make that field named "Uapi16ManifestFile" just because. It's the `contents` of the dir, and hence it should be named that way.

specs/file-manifest.md
poettering: I frankly not sure I fully grok the "forced" idea... can you elaborate what precisely shall be listed in it, and how it is used?

is "forced" supposed to be an allowlist of sorts, instead of the "revoked" marker that is kinda a denylist of sorts?

> One would have to have an entry for the version the client is on which might not be the case when it's a custom build. This requires the update server to be smart and be able to have the client version reported to inject such an entry marked as revoked.

maybe we should have the rule that any local versions that are not listed in the manifest should be considered revoked, or some state similar to it? i.e. anything not listed anymore means "yuck, not supported, move away". 

we could have three states basically for files:

1. listed in the manifest with revoked off or not set: GOOD
2. not listed in the manifest: BAD
3. listed in the manifest with revoked on: REVOKED

an updater should then make sure to always stay on an entry in state GOOD, and always get away from an entry BAD or REVOKED.

(it's not entirely clear to me how the policy should be different for BAD and REVOKED, maybe it should be treated the same even if we separate them otherwise)

poettering: i am really not a fan of the top-level additional arrays btw, because that means the file objects are not self-descriptive anymore: you cannot take them out of the manifest or they lose meaning. And this is a biggy I think, because I think for the varlink plugins for sysupdate we really want to take them out of the manifest and pass them individually over varlink IPC

poettering: i really don't grok the global tags idea I must say. Can you explain again? and should that really be part of the file manifest, or of some other data structure, possibly surrounding it?

specs/file-manifest.md
poettering: yes, this is key, we want the files objects to be self contained so that we can pass them around individually via IPC

specs/file-manifest.md
cyphar: @pothos 

Sorry, I was without internet and didn't realise my two comments got split -- my point was more that we should do priorities *and* they should be part of the per-file data.

I can see the argument for `forced` but it just feels a little too restrictive if we intend this to be a more general-purpose thing (and I think even for our needs it might be too restrictive).

It's also not entirely clear to me what the algorithm should be for picking versions with `forced` -- I guess the idea that every pattern that matches the file is forced to use that file? Since you need to do sorts based on version number, it seems to me that doing a two-key sort (priority, version number) is more akin to what sysupdate does anyway (though I might be wrong) and is more flexible. Having to work backwards from the forced files seems less ideal to me, but I could be wrong.

For the model we plan to have, priorities would also make a little more sense because (on a first pass) we would model the closest layers to the per-machine repo as the highest priority and then gradually lower the priority with each link. With forced we would need to then also compute which filenames should be forced, which would probably mean reimplementing sysupdate's selection logic to know which files need to be forced (and in the case of custom transfer files I suspect this is not really doable). If we instead treat priorities as an additional input to the sorting, we can keep the two things divided.

specs/file-manifest.md
poettering: i think keeping them separate is essential for security. 

i.e. let's say you want to replicate the full manifest tree locally for mirroring purposes. for that you probably want to copy all referenced objects unmodified (i.e. with their hashes intact) from the HTTP server to your storage. But for that you really want the original size of the object, so that you can abort the download/decoding if the download/decoding stream grows beyond the original object's size.

or in other words: knowing the size beforehand of what you will download matters, because you can abort the download if the size doesn't matter. It addresses the infinite download/compression explosion attacks.

of course, for sysupdate this does not matter, because it never truly mirrors, it only downloads the slice and authenticates it and is then happy, never bothering with anything more complete.

specs/file-manifest.md
poettering: you could just rewrite each entry, no? doing this a 2nd time on the whole manifest seems redundant to me?

specs/file-manifest.md
poettering: well, in a way https also can provide authentication. but sure we can look into adding this right-away.

specs/file-manifest.md
poettering: hmm, i think the version inherently needs to be in the name, because otherwise if you download resources sysupdate-style to some local place you'd lose the version there (since neither POSIX directories nor GPT partitions have a concept for that). and the version is not random metadata that is fine if lost during the download, because it's part of the identification of the resource.

hence I'd very clearly say: version must stay in the name

poettering:
so i wonder if we should maybe fold the stepping stone and revoked thing into a single "updatePolicy" field or so, giving that a revoked item should never be a stepping stone. And given this state is relevant only when using this manifests for sysupdate-style updaters I think it makes sense to make that clear in the name. Hence:

`updatePolicy` would be an enum with values `good`, `stepping-stone`, `revoked`:

 * `good` → the default for entries that are listed but do not have the field `updatePolicy` set. An updater should consider this entry as a great, regular update target.

 * `stepping-stone` → almost like `good`, but an updater should never skip over such an entry.

 * `revoked` → this is a tainted version, please upgrade asap, and downgrade if there's no upgrade
 
 * and then we could define one additional value `vanished` → this is a pseudo-state for items that where downloaded before but no longer appear in the manifest. In most contexts it should be treated just like `revoked`.

With this we'd have a single field encoding the whole policy, and it would be self-contained inside the file object (which is a property I like very much)

specs/file-manifest.md
pothos: It was meant as simple way to bypass the version selection logic (it should hold the *names* to install) but when we teach sysupdate to downgrade then one can either use 'revoked' or remove any newer versions, it's just more involved. Downgrading by default to the newest available version is a breaking change but I guess it's also quite niche and we can announce it in the release notes instead of having to make it opt-in.
But even with it being opt-in as long as we have a way to do it that's good enough for now and we don't need a global 'forced' then.

Doing something like priority 999 (or -999) on a file entry would also work, just needs scanning through all entries to make sure it's the highest.

pothos: We can also make this another file that sysupdate is able to fetch. The idea is that similar to what we have with *features*  for parts of a sysupdate component, we should have something that can instantiate a whole component from a template. Client side this would be a tags config entry in `/etc/` but to be able to manage a machine from the outside, a tags entry in the manifest here should be able to overwrite the local config. This way one can use confext/sysext with sysupdate for "GitOps", having a declarative machine configuration that says which extensions to run.

specs/file-manifest.md
pothos: Adding it in addition is very useful for processing this JSON file. Otherwise one has to parse the filenames and that requires knowledge of the match pattern.

specs/file-manifest.md
poettering: so we discussed elsewhere that we don#t want the "tags" concept in the manifest file anymore, but place it in a separate file, that encodes update policy in more detail. hence closing this comment here.

poettering: We discussed elsewhere now that we want to solve this update policy concept differently, and outside of the manifest, hence marking this conversation as resolved.

(background: the manifest generator and the policy manager are likely different results, and it makes sense to keep these things separate, i.e. store things in different files, so that they can be signed independently.)

specs/file-manifest.md
poettering: so i put the "any order" thing in there because sorting is hard and depending on the usecase you want different algorithms. i.e. usually one wants strcmp() style sorting i guess, except that for our case strverscmp() sorting would be much more useful. It seems wrong to me to mandate one ordering, because that would also imply defining in detail how that ordering should take place...

I will reword this a bit, explaining this in more detail. But I think the gist needs to be: "any order, but maybe think about it"

specs/file-manifest.md
AdrianVovk: I think the expiration date of an individual file is orthogonal to the expiration date of the manifest.

The manifest should expire to avoid a situation where an attacker serves a stale manifest w/ a stale list of files. Individual files can expire for any additional number of reasons: maybe the file is tied to a TPM secret that self-destructs by a certain deadline. Maybe we expire old versions of the image to make rollbacks harder. Maybe old images expire because we have certificates in there that expire. Really there's any number of reasons to expire an individual file. Point is: you may want to set a different expiration date on the manifest than the individual files

With the scheme you suggest, each time we update the manifest we need to:
- Separately keep a database of how the real expiration dates for each file
- Compute the expiration date of the manifest
- For each file set the expiration date to either the real expiration date or the manifest's expiration date, whichever comes first

So sure it's possible but it's a pain. We'd need separate storage for the real dates, and the manifest becomes less self-descriptive.

specs/file-manifest.md
AdrianVovk: > Detached signatures aren't too bad though, or? There are alternatives to GPG.

Detached signatures suck because they're separate files but need to be updated atomically as a pair. We've encountered issues where the SHA256SUMS and SHA256SUMS.gpg desync for the client, leading to signature check failures.

The leading hypothesis here is that our CDN is deciding to expire one of the files but not the other one, so there's a period of time where you might get a new manifest but an old signature (or vice versa). To fix this correctly, we'd need to figure out how to tell our CDN explicitly to expire both files together and re-fetch them. In practice, we just set a really short cache time-to-live for those two files and let the signature check fail occasionally for an unlucky client. By the time the client retries the update a while later, the files will probably be sync'd up again

If it's just one file, there's no more race condition to worry about

> in a way https also can provide authentication

Not nearly enough for our security model, though! An attacker could totally put sysupdate drop-ins in /run to point it at their own domain, which can use HTTPS without issue, and none of this would be measurable. The manifest must be validated against certificates that are measured into the TPM

specs/file-manifest.md
AdrianVovk: OK makes sense. I think we need to be clearer about that in the spec. i.e. that the `dataSize` describes the size of the decoded blob, _before_ a slice is applied

specs/file-manifest.md
AdrianVovk: > sysupdate-style to some local place you'd lose the version there (since neither POSIX directories nor GPT partitions have a concept for that)

While I generally agree that the version probably _should_ stay in the filename, I'd like to point out that this is talking about the download destination, which is going to have its own match pattern that will be used to name the destination file / partition. In other words: the download's filename has no bearing on the destination's filename

specs/file-manifest.md
```diff
@@ -0,0 +1,270 @@
...
+This is an object that encodes information about an individual file.
+
+```json
+{
+        "name" : "…",
+        "dataEncoding" : "gzip",
+        "encodedDataSize" : 2011,
+        "dataSize" : 4711,
+        "dataFile" : "…",
+        "dataUrl" : "http://…",
```
cyphar: For those without CDN budgets, it might be nice to allow specifying multiple URLs and the client can pick an arbitrary one?

poettering:
I pushed a new version, taking a lot of input into account.

I made three fundamental changes:

1. Instead of putting everything in a giant JSON array, I changed the thing to be a JSON-SEQ sequence instead. This takes @brauner's point regarding scalability into account: by making this JSON-SEQ the format can be read and processed in a stream fashion. if we use a giant json array this is typically not possible, most parsers would read the whole thing into memory. While that might be fine for certain usescases it certainly doesn't scaled to huge sizes. I opted for JSON-SEQ, since it's RFC specified and in contrast to other JSON sequence formats is fine with pretty printed JSON, and gives us clear markers where to put cut lines for cryptographic signatures.

2. the thing can now be used as a manifest for deeply nested file hierarchies, with full POSIX properties. Instead of handwavingly claiming that we can extend the format that way (as I did above) I now made the spec actually cover this fully, because it's not as trivial as I originally though.

3. The contents stuff now supports alternative sources as proposed by @cyphar. 

I think the outcome is quite nice and comprehensive. Of course JSON-SEQ means it is not just a text stream anymore, but I think the benefits outweigh the negatives here.

Anyway, please have a look.

I intend to put together some code in systemd to see how this feels when actually implementing this.

poettering:
I now prepped a patch for systemd's `systemd-dissect` tool (which already has --mtree), to generate a manifest from a directory tree. It looks really nice, and the patch is quite small actually:

https://github.com/systemd/systemd/pull/41990

Note that the output it generates is not really intended for systemd-sysupdate consumption. It includes uid/gid info after all (just to match the --mtree call), and for sysupdate that's unlikely what we want.

poettering:
btw, in case you wonder, jq can process json-seq with the --seq switch

poettering:
btw, for illustrative purposes, this is how a real-life /usr/include/ looks like in the currently described format:

https://paste.centos.org/view/raw/446a6f8e

it has dirs, regular files and symlinks.

I find that really readable with the naked human eye.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+is the size of a UAPI.16 file manifest generated for the subdirectory (which only applies if the directory
+contents are not provided in-line, but via a reference to a separate sub-manifest file, see above).
+
+The `major` (unsigned integer) field only applies to file objects of type `blk` and `chr`, it encodes the
+major device number of the device node. Similarly, `minor` (same type and condition) encodes the minor device
+number.
+
+The `mode` (unsigned integer) field encodes the UNIX access mode of the file object. It applies to all
+inode types, except `lnk`. Note that while UNIX access modes are typically written in octal, this one is
+encoded in a regular JSON number, i.e. decimal. The valid range is 0…4095 (i.e. `0o0000` to `0o7777`).
```
pothos, line 156: That's also how Ignition does it and with some tooling it's okay but when writing/reading this still is very strange to deal with. I wonder if a `0o` string wouldn't be the more natural embedding here.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+2. If `validAfterUSec` is set and greater than the current time, the download should immediately fail.
+3. If `validBeforeUSec` is set and lower than the current time, the download should immediately fail.
+4. The `contents` array should be iterated, and a suitable entry be picked as source for the file contents. If no `contents` array is specified, proceed as in step 7, with an implied `file` field set to the `name` of the file object.
+5. If the selected contents object has `literal` set, it should be Base64 decoded, continue in step 8.
+6. Otherwise, if `url` is set, the data from the URL should be acquired, continue in step 8.
+7. Otherwise, if `file` is set, the data should be acquired from the same URL as the manifest itself, however with the last component of the URL replaced by the file name encoded in `file`, following the same semantics as HTML relative links. Continue in step 8.
+8. If both `encodedSize` and `encoding` are set the size of the acquired data shall be checked against `encodedSize`.
+9. If `encodedSize` is not set, but `originalSize` is: the size of the acquired data shall be checked against `originalSize`.
```
pothos, line 357:
```suggestion
9. If `encodedSize` is not set, but `originalSize` is, the size of the acquired data shall be checked against `originalSize`.
```
Might match the previous sentence structure better

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+10. If `encoding` is set: the acquired data shall be decoded according to the algorithm indicated in `encoding`.
+11. If `encoding` and `originalSize` are set: the resulting decoded data shall be checked against `originalSize`.
```
pothos, line 359:
```suggestion
10. If `encoding` is set, the acquired data shall be decoded according to the algorithm indicated in `encoding`.
11. If `encoding` and `originalSize` are set, the resulting decoded data shall be checked against `originalSize`.
```

specs/file_manifest.md
pothos: Maybe `modeDec` to indicate decimal? :thinking:

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+
+When applied to a GPT partition table the first object may also describe fields in the overall GPT
+partition table header of the disk.
+
+Within each directory files may appear in any order. However, files nested into subdirectories must be listed
+immediately following the file object of the subdirectory object. (Even if ordering within directories is
+not mandated, it is definitely a good idea to sort the entries alphabetically, or by version
+sort. Alternatively, historical ordering – i.e. add new additions purely at the end – might be a good
+idea. Sorting entries ensures the manifests become reproducible as long as the same tool is used, which is
+definitely a welcome property.)
```
Foxboron, line 62: The parenthesis here encompassess too much text to be a note/sidenote thing. So preferably remove them or structure this another way?

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+
+The `name` fields of file objects must be uniquely assigned within a manifest file.
```
Foxboron, line 64: I think we can have singular here instead of plural? This harmonizes better with the next paragraph IMO.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+---
+
+# UAPI.16 File Manifest
+
+| Version | Changes         |
+|---------|-----------------|
+| 1.0     | Initial Release |
+
+This format stores information about static, immutable data resources that potentially can be acquired over
+the network or stored on a disk. It's in particular supposed to be able to reasonably represent the
```
keszybz, line 20: For the first sentence, "static" and "immutable" are both bogus. We have no control over those files, so whether they are static and immutable is completely out of scope.

The mtree man page says "The mtree format is a textual format that describes a collection of filesystem objects". That seems much more to the point. Maybe "This document describes a textual file manifest format that describes a collection of blobs. Those blobs can potentially be acquired over the network or stored on a disk. In particular, it can represent the properties of files arranged in a POSIX file system tree as well as partitions in a GPT partition table."

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+properties of files arranged in a POSIX file system tree, as well as partitions in a GPT partition table.
+
+This manifest file format is inspired by the traditional UNIX `SHA256SUMS` and BSD `mtree(5)` file formats,
+but has various additional features, among them:
+
+1. It allows referencing remote URLs as source for acquiring file contents.
+2. It allows extracting data "slices" from data sources, via offset and range.
+3. Various fields of additional per-file metadata may be defined, including various fields for GPT partition table metadata.
+4. It's an extensible format permitting vendors and projects to add their own per-file and per-manifest fields.
+5. Inline cryptographic signing.
```
keszybz, line 30: 1–4 are sentences with a subject, 5 should have one too.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+
+This format is designed with tools such as `mkosi` and `systemd-sysupdate` in mind, but it's intended to be
+generally useful as a way to describe collections of files that may potentially be acquired over the network
+or from local storage, and verified cryptographically.
```
keszybz, line 34: This repeats what is in the intro.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+## General Structure
+
+The file is organized as an [RFC7464 JSON-SEQ](https://datatracker.ietf.org/doc/html/rfc7464) sequence of
+file objects which are described below.
+
+The first object in the sequence always describes the enveloping directory. This *root* file object must
+*not* have a `name` field set (which identifies it as the *root* file object), however, it must have a
+`mediaType` field set (which identifies the sequence as a UAPI.16 manifest). File objects for directory
+entries `Uapi16Manifest` (or any filename with a prefix of `Uapi16Manifest.`), `.`, `..` or any path ending
+in `/.`, `/..` are not permitted.
```
keszybz, line 52: This second part should be a separate paragraph, since it describes a completely separate topic from the first part.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+
+When applied to a GPT partition table the first object may also describe fields in the overall GPT
+partition table header of the disk.
```
keszybz, line 55: Hmm, so the preceding para says that the first object must describe "a … directory". But then this sentence here contradicts that, and says that it may describe a GPT header.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+
+Within each directory files may appear in any order. However, files nested into subdirectories must be listed
+immediately following the file object of the subdirectory object. (Even if ordering within directories is
+not mandated, it is definitely a good idea to sort the entries alphabetically, or by version
```
keszybz, line 59: Please just use **must** and **should**, instead of "it's a good idea". E.g. "Files within a directory **should** be ordered alphabetically using [UAPI.10](https://uapi-group.org/specifications/specs/version_format_specification/) for version comparisons. In cases where files are added over time to the listing, a time-based ordering **may** be used."

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+This format is designed with tools such as `mkosi` and `systemd-sysupdate` in mind, but it's intended to be
+generally useful as a way to describe collections of files that may potentially be acquired over the network
+or from local storage, and verified cryptographically.
+
+## Manifest File Name and Media Type
+
+When stored in a file system directory – alongside the data files it references – the manifest file
+should be named `Uapi16Manifest`.
+
+When served via an HTTP server it's recommended to use the media type `application/vnd.uapi.16.manifest`.
```
keszybz, line 41: "This file format is identified by media type …"

(Basically, the part about HTTP server adds no value. In the example below this media type string is used, even though no HTTP is involved in any way.)

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+        "validAfterUSec": 4711,
+        "validBeforeUSec" : 4711,
+        "validity" : "revoked"
+}
+```
+
+### Fields
+
+All fields are optional. If a field shall not be set it can either be set to `null` or simply
+omitted in the JSON object, both ways shall be considered equivalent.
```
keszybz, line 118: This choice unfortunately breaks reproducibility…

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+
+If not otherwise indicated all fields are supposed to be of type string.
```
keszybz, line 120: ~~supposed to be~~

Maybe just "If not otherwise inidicated, fields are strings."

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+The `name` fields of file objects must be uniquely assigned within a manifest file.
+
+If the manifest encodes a GPT partition table, the `name` field must still be assigned for every file object, even
+if `gptLabel` is defined too. The former must still be unique within the manifest; the latter does not have to be.
+
+For manifests covering a directory tree with nested directories (as opposed to a flat directory listing
+without file objects of type `dir`), subdirectory listings may either be specified inline (typically
+preferable) or be imported via a separate "sub-manifest" file covering just this subdirectory (and potential
+sub-subdirectories, …), by referencing the separate UAPI.16 manifest file in the `contents` field of the file
+object.
```
Foxboron, line 73: I think all the paranthesis sectionts here can just be replaced by commas or rewritten to not require the paranthesis. 

Something liiiikkeee this maybe?

```
For manifests convering a direcotry tree with nested directories, as opposed to a flat direcotry listing without file
objects of type `dir`, it's preferably to specify the subdirectory listing inline or be imported bia a seperate
"sub-manifest" file covering just this subdirectory, by referencing the seperate UAPI.16 manifest file in the `contets`
field of the file object.
```

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+any other file object in the sequence.
+
+If `name` is not specified the record stores information about the top-level root file object. This file
+object always comes first in the file object sequence. Or in other words, a file object has exactly one of
+`mediaType` (if root file object/first object in sequence) and `name` (if not root file object/not first
+object in sequence) set.
+
+The `name` must be a valid UTF-8 POSIX path (i.e. using the slash `/` as component separator), may not
+contain control characters (ASCII 0…31, 127) and may not contain "`.`", "`..`", "``" as components of the
+path. It may not start nor end with a slash, nor may it have multiple slashes in immediate sequence (or in other
```
keszybz, line 133: "or"  (There are two valid forms: "It may neither start nor end with a slash" and "It may not start or end with a slash".)

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+defaults to `dir` for the first object in the sequence and `reg` for all others. Note that when using the
+manifest format to list partition contents each partition should be listed as a file of type `reg`.
+
+The `size` (unsigned integer) field only applies to file objects of type `reg`, `lnk`, `dir` and encodes the
+size of the contents of the file, in bytes. Specifically, for regular files this is the size as reported by
+the file system. In case of symlinks: this is the length of the symlink target string. In case of `dir` this
+is the size of a UAPI.16 file manifest generated for the subdirectory (which only applies if the directory
+contents are not provided in-line, but via a reference to a separate sub-manifest file, see above).
+
+The `major` (unsigned integer) field only applies to file objects of type `blk` and `chr`, it encodes the
```
keszybz, line 150: "only applies to" → "**can** only be specified for object of type  …

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+manifest file itself was found, `url` references a full URL, and `literal` contains in-line Base64
+data. `url` may use `http://…` and `https://…` schemes only. It is recommended (but not required) to use
+`literal` to encode the target of symbolic link file objects. If none of these fields are set, behaviour
+equivalent to `file` being set to the file object's `name` field should be implied.
+
+`literal` may be encoded either in regular Base64 or in URL-safe Base64. Consumers must be able to deal
+with either decoding vocabulary.
+
+`encoding` specifies the encoding of the specified resource. It accepts the same encoding specifiers as
+HTTP's `Content-Encoding:` field, i.e. `gzip` or `zstd`.
```
keszybz, line 274: "i.e." → "e.g." ?

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+and back.
+
+## Acquiring a File Remotely
+
+When acquiring a file listed in a UAPI.16 File Manifest from a web service, the following logic should be
+implemented.
+
+1. If `validity` is set to `revoked`, the download should immediately fail.
+2. If `validAfterUSec` is set and greater than the current time, the download should immediately fail.
+3. If `validBeforeUSec` is set and lower than the current time, the download should immediately fail.
```
keszybz, line 351: Hmm, why not "since" and "until"? "after" and "before" as described here doesn't match the description: "after"/"before" implies that the boundaries are excluded, while the description specifies that they are included in the valid range. In practice this doesn't matter much, but is inelegant.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+All fields are optional. If a field shall not be set it can either be set to `null` or simply
+omitted in the JSON object, both ways shall be considered equivalent.
+
+If not otherwise indicated all fields are supposed to be of type string.
+
+The `mediaType` field is mandatory on the first file object in the sequence (i.e. the top-level root
+directory object), and must have the value `"application/vnd.uapi.16.manifest"`. It should not be used on
+any other file object in the sequence.
+
+If `name` is not specified the record stores information about the top-level root file object. This file
```
keszybz, line 126: This says that `name` is first, but I don't see the order of other entries described anywhere.

keszybz (commented):
I think that the primary problem with the current text is that it aims for reproducibility, but leaves a bunch of implementation choices undefined:
- sorting of entries
- sorting within an entry
- null or omitted fields
- whitespace formatting of json

specs/file_manifest.md
keszybz: I think this should be reworded to say that the first object identifies the file format as a whole and may be describe either a directory or a GPT partition header.

Foxboron (commented):
LGTM. But generally I think there is an over-reliance on paranthesis for what should be properly part of the paragraphs.

Foxboron:
@keszybz The reproducible builds aspect is a good catch. The file listing should be ordered in some declared way. Probably alphabetical order?

cyphar:
As the new draft is quite amenable to outputting streamed data, I wonder if there should be a special "end of list" record to indicate that a manifest is complete so that if a producer crashes or errors out consumers won't incorrectly consume the partially-output manifest.

This does cause some issues with signature objects, as ideally you'd want them to be put after the "end of list" record but then they need special handling. And if you don't do that then you can take a signed manifest, put stuff after the signature and hope that consumers make mistakes parsing them.

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+{
+        "mediaType" : "application/vnd.uapi.16.signature",
+        "signatures" : [{
+            "mechanism" : "openpgp",
+            "key" : "38110B0FAC93C324",
+            "data" : "iQIzBAABCgAdFiEEFrHE7sC8AhrHd/aBtjshh5w0hbAFAmn8rUoACgkQtjshh5w0hbAz3g//eyWO1swKXivio1PH2oJclEAYqpyqkv9LC293ePyDp5K0zL7BEVt9ijCGb22Wcu+xBXug4SBpSsCMOcro633Kn0ACBeOHwsyu1rnumb85/xvx3lSiq7ke1HcP5OWh5XX4lTm3KYVaPLQy0P3ThupJJsCa0o7uDdl2fXEgcZtRGhgp7GnpQabEuUUGW+1YZ1DFhhIdewZ6+eZZac0DPFjevBbI14jf9wEp3uWT9YoULF0wWBUi45vxdievwIRbcQHQVhxKMaTm0yuTbT5kSUAZccsOkJ4gqrVoNmFx0kYPve0uWcEkFQ16zEKVESr/fhDuCW19v3D46JNvwxZgIxP3o5rp3wbOja6n5D8Otl0kUQaE7FPOmYI1SZ2F97cYsKNViuUyrMcPTNGkI4E18H4WgWLf1CEyGHT16KmoCmldqVhQiTFCtOzkJdbwJgnSL/f0n2y3H/iB3lL0hglv2SDDVnVeMBx+qs+j2DMys3SVOOAYqGrMay1PYxLTEct+M1D0aAmH+XObuoVcm3XMtPPt6NCzxes0cdmFCdMTGERYLY522YCruJpO7p4h9IEoKqDACot0vDD0qwnjBCPu7ETrb02VvJ/W3GJ7/3Y8lIvZFtJqs5TQuxU0d4OiDHe08XdwZrcAondrX5h3Up7mpQIIF4RAMqUaYdOWEf7WzFZBKvY="
+        }]
+}
+```
+The above provides three separate files `FooOS.raw`, `FooOS-esp.raw`, `FooOS-root.raw`. All files are backed by the same data file. The latter two are
```
Skyb0rg007, line 473: This explanation is very confusing -- if I understand correctly there is one gzipped file containing a GPT partition, and that GPT partition has (at least) two partitions: a root partition and an ESP partition.
But then it seems like the manifest is asking the downloader to copy the ESP and root partitions out of the GPT image, but also to keep the original file containing those partitions too?

specs/file_manifest.md
```diff
@@ -0,0 +1,475 @@
...
+
+```json
+{
+        "file" : "…",
+        "url" : "https://…",
+        "literal" : "TmV2ZXIgZ29ubmEgZ2l2ZSB5b3UgdXAsIG5ldmVyIGdvbm5hIGxldCB5b3UgZG93bg==",
+        "encoding" : "gzip",
+        "encodedSize" : 2011,
+        "originalSize" : 4711,
+        "offset" : 22
```
Skyb0rg007, line 255: Add `"sha256"` for completeness, as it lets the client bail before starting to decompress any data.

specs/file_manifest.md
bdrung, line 41: Since this format is based on JSON-SEQ, the media type should be ``application/vnd.uapi.16.manifest+json-seq` then.

specs/file_manifest.md
poettering: well, it's not decimal, it's just json native integer...

specs/file_manifest.md
poettering: The reason I put "static" an "immutable" in, is that the manifest encodes a hash of the data. Hence, the data cannot change and cannot really be generated dynamic unless reproducibly so, I'd call that static and immutable.

specs/file_manifest.md
poettering: yupp, the whole design assumes that json is not reproducible among implementations. because it really isn't. every implementation formats things differently.

i think reproducibility is great for this, if you stick to one tool in one version, but expecting reproducibility across implementations I think is not realistic, already sd-json python and rust/serde would differ how they'd format certain things.

accepting this is the reason why the signing mechanism describe here is about the formatted data preceeding in the json-seq sequence, without any attempt requiring normalization.

specs/file_manifest.md
poettering: well, there are at least three different sorting schemes I find acceptable:

1. strictly alphabetical
2. strictly by version sort
3. time-based: later added stuff is at the end

declaring a SHOULD here seems a bit weird to me, given how many different orders I think should be acceptable here. In particular as sorting unicode correctly and reproducibly is kinda hard...

hence my choice of vague language: i think we should say that everything is ok, but sticking to an order is a good idea.

specs/file_manifest.md
poettering: hmm, there's a whole discussion of the ordering further up?

poettering:
> As the new draft is quite amenable to outputting streamed data, I wonder if there should be a special "end of list" record to indicate that a manifest is complete so that if a producer crashes or errors out consumers won't incorrectly consume the partially-output manifest.

> This does cause some issues with signature objects, as ideally you'd want them to be put after the "end of list" record but then they need special handling. And if you don't do that then you can take a signed manifest, put stuff after the signature and hope that consumers make mistakes parsing them.

One thing I was thinking of is whether we should put a trailer record after the file records, always, carrying just the hash of what comes before. Would basically separate the file entry objects from the signature objects.

or alternatively, add a pseudo-signature object as an option: one that doesn't contain any real signature but does contain a hash, and may be listed among the signature objects, or alone.

specs/file_manifest.md
poettering: the gzipped data (and the first defined file) covers the whole disk image. the two other files cover two partitions that are part of the disk image. Since these partitions will mostly contain the same data as the whole disk image we only put up a single source file, that clients can extract what they need from.

will reword this a bit.

specs/file_manifest.md
poettering: > Since this format is based on JSON-SEQ, the media type should be ``application/vnd.uapi.16.manifest+json-seq` then.

can you explain this? why would the bet better/more appropriate?

poettering:
I updated the spec in various ways, adding sections about extensibility and reproducibility.

i addressed most of the points raised above, except for the ones i left open

specs/file_manifest.md
poettering: added a section about reproducibility

specs/file_manifest.md
bdrung: The file format is a json-seq file. Adding `+json-seq` will allow application that support json-seq to open the file.

cyphar:
> or alternatively, add a pseudo-signature object as an option: one that doesn't contain any real signature but does contain a hash, and may be listed among the signature objects, or alone.

I think this is a pretty nice idea actually (it nicely avoids the need for another special node and makes more semantic sense). The main downside is kind of [the issue I referenced above](https://github.com/uapi-group/specifications/pull/213#issuecomment-5335896190) -- if we allow signatures in the middle of the stream a consumer needs to be very careful to make sure that they partition their trust of the manifest based on where the signature appears. (In other words, make sure that an entry at the end with no signature after it doesn't get treated as though it had been signed by an earlier signature, as well as making sure if only certain signatures can sign for files -- basically making sure that you correctly associate which signatures are authenticating which entries.)

In practice I reckon that means you'd want to parse the stream in reverse order to make it harder to have bugs of that kind, as well as to check for the "end of records" empty signature object before consuming the manifest data and potentially doing something silly based on it.

- - -

If you'd prefer to leave this undefined for now, we can have people implement their own takes on this and then revisit this after publishing?

poettering:
I discussed this for a while with @brauner today. I think the ability to append something has little real usecase (because append-only sucks, because access times would suck if you don't rotate, because you need to auth the whole manifest each time, but if you rotate then ability to append isn't that useful anymore, and having "partial" sigs also complicates implementation, which sounds risky).

So I think there's value in having a trailer in place that indicates completion. Christian's and my last idea we found prettiest would be require the trailer after the last signature, i.e. at the very final end of the json-seq stream. and it would be a json object with a media type of `application/vnd.uapi.16.trailer` which is otherwise empty (at least for now).

Then we'd basically have the general structure:

1. top-level intro object for the root inode (exactly one)
2. entry objects for the contained inodes (zero or more)
3. a series of signature objects (zero or more)
4. a final trailer object

and future extensions would be permitted to introduce objects anywhere in the stream, execept before the intro object or after the trailer. if they place it before the first sig it would be protected, if after it it would not be.

I think this is quite nice to communicate and very simple, no?

We discussed for a while if the boundary between entry and signature objects should be some explicit boundary object (maybe marked with application/vnd.uapi.16.boundary media type or so), but we decided against that, it would not really bring any benefit, and putting the boundary between two objects (i.e. on a byte boundary) instead of on an object seems like the simpler, better choice.

does that make sense? if so, i'd update the spec accordingly.

poettering:
I added the trailer object to the spec now.

septatrix:
I think providing a JSON schema for this would be great to easily validate files

cyphar:
> and future extensions would be permitted to introduce objects anywhere in the stream, execept before the intro object or after the trailer. if they place it before the first sig it would be protected, if after it it would not be.

I'm not sure that you'd want extensions to put things after the signed section for the same "what does the signature protect" problem I mentioned above. I think you'd always want signatures to be final, which means that the no-signature variant is a little cleaner IMHO.

PR #218, opened 2026-06-17
Rework the descriptions of directories for shared libraries and the symlinks to them

keszybz:
The first commit was the initial motivation: the description of multiarch and multilib was edited too many times and was rather confusing to read. It also contained some incorrect statements. The other two commits are cleanups for the surrounding text. Each has an extensive description in the commit message, not repeated here.

### DIFF
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -264,7 +264,6 @@ On "multilib" systems,
 a simpler scheme is used that only supports 32-bit and 64-bit variants of the same architecture.
 
 Both schemes are widely used, and distributions typically choose one or the other.
-Multiarch is the recommended approach, especially for new systems.
 
 Note that many locations described in this section
 are under shared ownership,
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -514,17 +514,18 @@ correctly find their binaries.
 
 ### `/lib/`
 
-This compatibility symlink points to `/usr/lib/`,
-ensuring that programs referencing this legacy path
-correctly find their resources.
+A symlink required for the correct execution of binaries specifying
+the dynamic loader with a path starting with `/lib/`.
+This symlink points to `/usr/lib/`.
 
 ### `/lib64/`
 
-On some architecture ABIs, this compatibility symlink points to `$libdir`,
-ensuring that binaries referencing this legacy path
-correctly find their dynamic loader.
-This symlink only exists on architectures whose ABI
-places the dynamic loader in this path.
+A symlink required for the correct execution of binaries specifying
+the dynamic loader with a path starting with `/lib64/`.
+The directory to which it points varies between multiarch and multilib installations
+and even between different distributions using the multilib layout.
+This symlink may not exist on architectures whose ABI
+does not place the dynamic loader in this path.
 
 ### `/var/run/`
 
--- a/specs/linux_file_system_hierarchy.md
+++ b/specs/linux_file_system_hierarchy.md
@@ -303,12 +303,23 @@ This layout supports parallel installation of 32-bit and 64-bit variants of the 
 
 This directory contains architecture-independent files,
 internal binaries and other programs that are not regularly invoked from a shell,
-and shared libraries for the 32-bit architecture.
+and shared libraries for either the 32-bit or 64-bit architecture.
 
-##### `/usr/lib64/`
+##### `/usr/lib64/` or `/usr/lib32/`
 
-This directory contains shared libraries for the 64-bit architecture.
-It can be used for architecture-dependent package-specific data too.
+This directory contains shared libraries for the other architecture.
+
+Either directory can be used for architecture-dependent package-specific data too.
+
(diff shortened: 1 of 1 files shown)


bluca (approved):

keszybz:
There's one more commit from @behrmann to describe what happens in Arch. PTAL.

bluca (approved):

Foxboron (approved):

gotmax23:
The are still some places in the doc that only mention the Debian multiarch layout. Search for `<arch-id>`.

Also, in the newly added section, should the section for the more common multilib layout go first?

keszybz:
> The are still some places in the doc that only mention the Debian multiarch layout. Search for `<arch-id>`.

Indeed. I adjusted the text in three locations.

> Also, in the newly added section, should the section for the more common multilib layout go first?

I think it's fine as-is. We are explicitly not saying that one is better than the other, so the current alphabetical order is fine.

poettering:
i am not convinced about much of this, this just describes the status quo, but does not provide guidelines for the future, for doing it right. While that might be interesting to some i don't think it's useful for a spec. it's descriptive not prescriptive.

i.e. what i am getting from this is that this makes lib32 and lib64 a recommended thing, pretends it was a pinnacle of good design. although it's actually just a hack if you ask me. lib64 should exist for compat with the dynamic loader, but otherwise the only reason why anyone would have that would be compat with traditional fedora-based distros, but that doesn't have to be in the spec, people can just add that on their own implementations.

the original spec made clear that extensions where allowed in particular for compat and lib32/lib64 could be considered covered by that. I am really not convinced by lib32/lib64 deserves any recommendation here.

also a spec that says anything goes (i.e. blesses both multilib/multiarch, without judgement) might as well not exist, because it just describes, it doesn't mandate or recommend anything.

maybe split this up? i.e. one descriptive "research" doc that just describes what fedora, suse, debian, arch do, and a separate one that makes a prescriptive recommendation like the original spec did? then i can ignore the research doc, and shut up about complaining about it?

poettering:
i mean, the libexec/ stuff is similar: there's really no point in having it. fedora has it, i think it should just die because it doesn't bring any benefits. the spec at leasts says that it should not be used. i think lib64 is very similar: it's not a good design. we should mandate the /lib64/ symlink but otherwise not recommend it as anything, because it's a mess.

behrmann:
> also a spec that says anything goes (i.e. blesses both multilib/multiarch, without judgement) might as well not exist

The changes are mostly descriptive because there were a few people who asked for that and, if I remember correctly, also asked why multiarch was prescribed. 

I have little preference there, but recommending multiarch seems… in lack of a better term, untimely? Debian adopted multiarch around 2012, if I remember correctly, and I don't think many distributions have sprung to adopt it since, so there is Debian world with multiarch and the rest. multiarch does seem like the better design, but it lacks adoption outside of Debian land and I don't think recommending it here will change that. 

I think the question is: What to recommend? Is recommending multiarch still timely or would this be approached differently today?

keszybz:
> maybe split this up? i.e. one descriptive "research" doc that just describes what fedora, suse, debian, arch do, and a separate one that makes a prescriptive recommendation like the original spec did?

I seriously considered this now, but I don't think it's a great idea. For two reasons. First, I want people to use this spec and take is seriously in packaging guidelines and such. This means that it must describe a practical system. If we remove any discussion of multiarch/multilib, the utility of the text goes way down. In practice, *all distributions* cater to mixed-arch installations. If the text is split, then people have to say "uapi.9 + the second document that actually describes what we do" and nobody wants to do that. Second, the parts that would be split out is `/opt`, `libexec`, and the multiarch/multilib stuff, and that is a handful of paragraphs out of a very long text. This really seems like a waste of work to do such a split.

> we should mandate the /lib64/ symlink but otherwise not recommend it as anything, because it's a mess.

This is what the text did *before* this commit and this is what the text is doing *after* this commit. The change was to fix the statement that it should be linked to $libdir, which is completely wrong, as described in the commit message.

> libexec/ stuff is similar: there's really no point in having it. fedora has it, i think it should just die because it doesn't bring any benefits. the spec at leasts says that it should not be used.

There are two categories of "unneeded things": things that are actively broken and things that are more in the bikeshed category. For example `/opt` belongs in the former category, because of lack of integration with the system. OTOH, `libexec` is mostly in the latter. There is no practical difference between `/usr/lib/mypackage/mypackage-helper` and `/usr/lib/libexec/mypackage-helper`, because everything in `/usr/lib`/ is stored on the same partition with the same access properties. It it *strictly* a matter of preference. Anyway, the spec labels `libexec` with ⚠️, so we don't need to rehash this.

> also a spec that says anything goes (i.e. blesses both multilib/multiarch, without judgement) might as well not exist, because it just describes, it doesn't mandate or recommend anything.

In practice people are going to follow the layout that their distro has. Changing from one to the other is an insane amount of work. If you want, we could recommend one of the variants. Previously we recommended the multiarch layout, but didn't even get its description right. Fedora-style multilib is the most popular. IMO the Arch-style multilib is more elegant…

lnussel:
What is the target audience of the document?

PR #219, opened 2026-06-29
tpm-pcr-registry: document systemd extending PCR 7 with "os-separator"

keszybz:
Added to systemd 261 in
https://github.com/systemd/systemd/pull/41016/changes/7c27f9f59455b199e12976c76b28b708c525b55c#diff-e185456c303f7a77bef6607af135857668f4b30b67f21ea372d45f7d69927730R24.

### DIFF
--- a/specs/linux_tpm_pcr_registry.md
+++ b/specs/linux_tpm_pcr_registry.md
@@ -91,11 +91,29 @@ In both cases it is important that data measured into the PCRs is carefully chos
   </tr>
 
   <tr>
+   <td class=number></td>
+   <td class=systemd><code>systemd-pcrosseparator.service 🚀</code></td>
+   <td>Userspace</td>
+   <td><code>os-separator</code> barrier when leaving initrd</td>
+   <td><code>/run/log/systemd/tpm2-measure.log</code></td>
+   <td>n/a</td>
+  </tr>
+
+  <tr>
    <td class=number><p style="text-align: right"><strong>1</strong></p></td>
    <td class=firmware><code>Firmware 💻</code></td>
    <td>UEFI Boot Component</td>
    <td>Core system firmware data/host platform configuration; typically contains serial and model numbers</td>
    <td>UEFI TPM event log</td>
+   <td>n/a</td>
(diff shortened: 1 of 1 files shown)


Foxboron (approved):

specs/linux_tpm_pcr_registry.md
```diff
@@ -294,6 +402,9 @@ PCR 5 changes when partitions are added, modified, or removed.
 
 PCR 7 changes when UEFI SecureBoot mode is enabled/disabled, or firmware certificates (PK, KEK, db, dbx, …) are updated.
 The shim project will measure most of its (non-MOK) certificates and SBAT data into this PCR.
+[systemd-pcrosseparator.service](https://www.freedesktop.org/software/systemd/man/systemd-pcrosseparator.service.html)
+extends PCRs 0–7, 9, 12–14 with a "<code>os-separator</code>" barrier when leaving the initrd,
```
behrmann: The table is HTML because Markdown can't do tables this complicated otherwise, but here it's not necessary:
```suggestion
extends PCRs 0–7, 9, 12–14 with a `os-separator` barrier when leaving the initrd,
```
The quotes around `os-separator` also look a bit weird in the table, when rendered. Since the thing is already rendered as code I think those can be dropped.

Looking at the docs in the linked commit, this could be a bit clearer as
```suggestion
extends PCRs 0–7, 9, 12–14 with the literal word `os-separator` as a barrier when leaving the initrd,
```

keszybz: Updated with all the suggested changes.

behrmann (approved):

poettering:
ah, uh, sorry i forgot to add that.

but note that "leaving" is not entirely precise here. It's invoked after the tpm device pops up, but we do not wait any longer than that. So it might not be a the beginning of the initrd (since tpm might take a bit to pop up), but it's definitely more the beginning than the end, since we do not explicitly delay this.

poettering:
→ https://github.com/uapi-group/specifications/pull/221

PR #222, opened 2026-07-07
specs: document ELF note-merging semantics for dlopen and package metadata

xnox:
Linkers concatenate same-named sections from multiple input files, so a
linked binary can carry multiple .note.dlopen or .note.package notes even
when each producer emitted only one. Parsers must therefore iterate over
one or more Note structures rather than assuming a single note.

Add a "Guidance on parsing ELF sections" section to both specs describing
this, with a readelf example. Make the two specs share the same parsing
procedure: this behavior is forced by the ELF spec and by the binutils ld
and LLVM lld implementations, and was already the long-standing status quo
for dlopen metadata (one note per dependency or one combined note), so the
package metadata spec now inherits it.

For package metadata, note the SCA/SBOM/vulnerability-matching implication:
multiple notes indicate a likely statically linked binary composed from
several packages, all of which must be enumerated.

### DIFF
--- a/specs/elf_dlopen_metadata.md
+++ b/specs/elf_dlopen_metadata.md
@@ -75,6 +75,10 @@ Strings shall not contain any control characters or use `\uXXX` escaping.
 Reference implementations of [packaging tools for `.deb` and `.rpm`](https://github.com/systemd/package-notes)
 are available, and provide macros/helpers to parse the note when building packages and adding dependencies.
 
+Note section is specified in the [Tool Interface Standard (TIS) Executable and Linking Format (ELF) Specification version 1.2](https://refspecs.linuxfoundation.org/elf/elf.pdf).
+
+Linkers may merge multiple note sections; see [Guidance on parsing ELF sections](#guidance-on-parsing-elf-sections) below.
+
 ## Well-known keys
 
 The metadata format is intentionally extensible, so that upstreams and later revisions of this spec can add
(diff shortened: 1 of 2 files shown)


specs/package_metadata_for_executable_files.md
```diff
@@ -246,3 +250,37 @@ $  coredumpctl info
 ```
 
 (This is just a simulation. `ls` is not prone to crashing with a segmentation violation.)
+
+## Guidance on parsing ELF sections
+
+When ELF linkers encounter an identical section name across multiple input files, the sections are concatenated in the order they appear, padded and aligned. This merging is standard linker behavior and is generally not under the control of the tool that emits the note. See the [binutils ld](https://sourceware.org/binutils/docs/ld/SECTIONS.html) documentation as well as [LLVM lld](https://releases.llvm.org/22.1.0/tools/lld/docs/ELF/linker_script.html#linker-script-implementation-notes-and-policy), which describe the same behavior.
```
poettering: padded/aligned to what?

specs/elf_dlopen_metadata.md
```diff
@@ -198,3 +202,75 @@ $ dlopen-notes --rpm-requires archive --rpm-recommends bpf  /usr/lib64/systemd/l
 Requires: libarchive.so.13()(64bit)
 Recommends: libbpf.so.1()(64bit)
 ```
+
+## Guidance on parsing ELF sections
+
+When ELF linkers encounter an identical section name across multiple input files, the sections are concatenated in the order they appear, padded and aligned. This merging is standard linker behavior and is generally not under the control of the tool that emits the note. See the [binutils ld](https://sourceware.org/binutils/docs/ld/SECTIONS.html) documentation as well as [LLVM lld](https://releases.llvm.org/22.1.0/tools/lld/docs/ELF/linker_script.html#linker-script-implementation-notes-and-policy), which describe the same behavior.
```
poettering: padded/aligned to what?

poettering:
lgtm, but be more precise regarding the padding/alignment, as suggested

specs/package_metadata_for_executable_files.md
xnox: it is a bit convoluted - it is 4 bytes; but also the max bytes of all sections being merged if different alignment specified, and could be 8.

will collect precise wording from the ELF spec and ld implementation to be precise & concise.

Will be able to get to this in a weeks time.

specs/elf_dlopen_metadata.md
xnox: will update in a week

specs/package_metadata_for_executable_files.md
xnox: it is 4 bytes actually everywhere. So just adding that.

specs/elf_dlopen_metadata.md
xnox: it is 4 bytes actually everywhere. So just adding that.

PR #223, opened 2026-08-19
Add UAPI.18: CLI introspection

behrmann:
This is a draft that @keszybz based his work in https://github.com/systemd/systemd/pull/43152 on (with a few typos) fixed. A few commits have been put on top addressing a few things that where found during the review of that PR already and that were commented on https://github.com/behrmann/uapi-specifications/commit/e7911dc962f2c7bedf6bc8a0218912b86e07f3af#comments. It also adds what's currently pending in https://github.com/systemd/systemd/pull/43430. 

The idea is to have a JSON schema that describes the full output (or a superset of that) that one would get from a binary via `--help` in a structured way. The idea is that this would allow to automatically generate command line completion from this, which we have already done in mkosi, but would also allow other things, like checking for features of a command line program without resorting to brittle parsing of help output.

This is still very much work in progress.

### DIFF
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -274,15 +274,27 @@ either to a positional argument or an optional argument.
 }
 ```
 
-All keys except `type` and one of `value`, `dynamic` or `missing` are optional
-and are treated as empty string or false when missing.
+All keys except `type` and one of `value`, `category`, `dynamic` or `missing`
+are optional and are treated as empty string or false when missing.
 
 `type` is the fixed string `value` and signals that this is a value object,
 describing an argument value.
 
-`value` is a string describing a possible value.
+`value` is a string describing a possible static value.
 
-`dynamic` is a non-empty array of strings describing a sequence of commands,
+`category` is a string describing a category of values:
+- `path`, any filesystem path,
+- `file`, a filesystem path to a regular file,
+- `dir`, a filesystem path to a directory,
+- `pid`, a PID,
+- `uid`, a numeric UID,
+- `gid`, a numeric GID,
+- `username`, a user's name,
+- `groupname`, a group's name,
+- `hostname`, a hosts' name, and
+- `unit`, a systemd unit's name.
+
+`dynamic` is a non-empty string describing a commands,
 that can be called to generate multiple values.
 This is relevant for the dynamic generation of completion candidates during
 command line completion.
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -148,6 +148,7 @@ Value objects describe a value that is passed as an argument, either to a
     "help": "help text",
     "default": True
 }
+```
 
 All keys except `name` are optional and are treated as empty string or false
 when missing.
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -69,7 +69,7 @@ The command object is recursively defined as follows.
     "options": [<option object>, …],
     "verbs": [<command >, …],
     "documentation": ["doc1", …],
-    "package": "myproject"
+    "project": "myproject"
 }
 ```
 
@@ -102,7 +102,7 @@ that should not be larger than 16.
 `documentation` is an array of string values describing URIs referencing
 documentation for this command, see `man:uri(7)`for a description of valid URIs.
 
-`package` is a string describing what this command belongs to. This may the
+`project` is a string describing what this command belongs to. This may the
 package that installed it or the project that produced it.
 
 ### Argspec objects
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -272,18 +272,36 @@ either to a positional argument or an optional argument.
 }
 ```
 
-All keys except `type` and `name` are optional and are treated as empty string
-or false when missing.
+All keys except `type` and one of `value`, `dynamic` or `missing` are optional
+and are treated as empty string or false when missing.
 
 `type` is the fixed string `value` and signals that this is a value object,
 describing an argument value.
 
 `value` is a string describing a possible value.
 
+`dynamic` is a non-empty array of strings describing a sequence of commands,
+that can be called to generate multiple values.
+This is relevant for the dynamic generation of completion candidates during
+command line completion.
+The current input for the argument,
+if any otherwise an empty string,
+will be passed as first and only argument to the first element of the array and
+its standard output stream will be passed as input stream to the next element
+and so on with the standard output stream of the last element defining the
+values,
+one per line.
+
+`missing` is a boolean signalling that some values are missing from the
+description.
+
 `help` is a string describing the help text that should be shown for the value.
 
 `default` is a boolean describing whether this value is the default value for
 the option this is a value for.
+
+If multiple of `value`, `dynamic` and `missing` are defined,
+`missing` has the highest precedence, followed by `value` and `dynamic`.
 
 ## Example
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -67,7 +67,7 @@ The command object is recursively defined as follows.
     "abstract": ["paragraph1", "paragraph2", …],
     "postscript": ["paragraph1", "paragraph2", …],
     "options": [<option object>, …],
-    "verbs" : [<command >, …]
+    "verbs": [<command >, …]
 }
 ```
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -67,7 +67,8 @@ The command object is recursively defined as follows.
     "abstract": ["paragraph1", "paragraph2", …],
     "postscript": ["paragraph1", "paragraph2", …],
     "options": [<option object>, …],
-    "verbs": [<command >, …]
+    "verbs": [<command >, …],
+    "documentation": ["doc1", …]
 }
 ```
 
@@ -96,6 +97,9 @@ arrays may be empty. Each element of the array represents a paragraph of text.
 `verbs` are command objects that describe a program's verbs, also called
 subcommands. Verbs may recursively define further verbs up to a maximum depth
 that should not be larger than 16.
+
+`documentation` is an array of string values describing URIs referencing
+documentation for this command, see `man:uri(7)`for a description of valid URIs.
 
 ### Argspec objects
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -287,11 +287,9 @@ that can be called to generate multiple values.
 This is relevant for the dynamic generation of completion candidates during
 command line completion.
 The current input for the argument,
-if any otherwise an empty string,
-will be passed as first and only argument to the first element of the array and
-its standard output stream will be passed as input stream to the next element
-and so on with the standard output stream of the last element defining the
-values,
+if any,
+otherwise an empty string will be passed as first and only argument.
+The standard output stream of the command defines the values,
 one per line.
 
 `missing` is a boolean signalling that some values are missing from the
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -165,7 +165,10 @@ long-options. Short options are usually followed by a single chracters, whereas
 long options can be a longer string. Short otions may be followed by multiple
 characters, but this is discouraged.
 
-`argument` defines the argument type TODO
+`argument` defines the argument type, which is one of the strings:
+-`no_argument`,
+-`required_argument`, or
+-`optional_argument`.
 
 `help` is a string that defines the help text of the option.
 
@@ -187,7 +190,7 @@ Value objects describe a value that is passed as an argument, either to a
 {
     "name": "myname",
     "help": "help text",
-    "default": True
+    "default": true
 }
 ```
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -36,13 +36,13 @@ The schema is defines as
 
 ```json
 {
-    "mediaType" : "application/vnd.io.systemd.cli-introspection-0",
+    "mediaType" : "application/vnd.uapi-group.cli-introspection-0",
     "commands": [<command object>, …]
 }
 ```
 
 where `mediaType` is the fixed string
-`application/vnd.io.systemd.cli-introspection-0` and `commands` is a non-empty
+`application/vnd.uapi-group.cli-introspection-0` and `commands` is a non-empty
 JSON array of *command objects*.
 
 Further additions to this specification will only be additive and non-backwards
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -194,7 +194,8 @@ This is only for display-purposes.
 
 `values` is an array of *value objects* describing the values this option may
 take.
-An empty array describes an argument that is a single arbitrary word.
+An empty array describes an argument that is a single arbitrary word with no
+further documented semantic.
 Value objects are described in a section below.
 
 `isDeprecated` is a boolean describing whether this option has been deprecated.
@@ -252,7 +253,8 @@ This is only for display-purposes.
 
 `values` is an array of *value objects* describing the values this option may
 take.
-An empty array describes an argument that is a single arbitrary word.
+An empty array describes an argument that is a single arbitrary word with no
+further documented semantic.
 Value objects are described in a section below.
 
 `isDeprecated` is a boolean describing whether this option has been deprecated.
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -69,7 +69,7 @@ The command object is recursively defined as follows.
     "options": [<option object>, …],
     "verbs": [<command >, …],
     "documentation": ["doc1", …],
-    `package`: "myproject"
+    "package": "myproject"
 }
 ```
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -156,7 +156,8 @@ explicitly made.
     "help": "Show this help",
     "group": [""],
     "conflicts": [],
-    "values": [<value object>]
+    "values": [<value object>],
+    "section": "option block 1"
 }
 ```
 
@@ -188,6 +189,8 @@ this option may not be called together with.
 
 `values` is an array of *value objects* describing the values this option may
 take. Value objects are described in a section below.
+
+`section` is a string to visually group options into separate lists.
 
 ### Value objects
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -70,6 +70,7 @@ The command object is recursively defined as follows.
     "features": ["feature1", "feature2", …]
     "abstract": ["paragraph1", "paragraph2", …],
     "postscript": ["paragraph1", "paragraph2", …],
+    "help": "short help"
     "options": [<option object>, …],
     "verbs": [<command >, …],
     "documentation": ["doc1", …],
@@ -97,6 +98,10 @@ program.
 `abstract` and `postscript` define arrays of strings that are shown respectively
 before and after the description of options in the program's help output. Each
 element of the array represents a paragraph of text.
+
+`help` defines the help text of the command. This is useful for single-line
+usage information as well as for short descriptions of verbs in the list of
+options.
 
 `options` define a CLI's positional arguments and options and are described in a section below.
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -85,7 +85,7 @@ builtin-in features of this program.
 
 `abstract` and `postscript` define arrays of strings that are shown respectively
 before and after the description of options in the program's help output. These
-arrays may be empty.
+arrays may be empty. Each element of the array represents a paragraph of text.
 
 `options` define a program's arguments and options and are described in a section below.
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -68,7 +68,8 @@ The command object is recursively defined as follows.
     "postscript": ["paragraph1", "paragraph2", …],
     "options": [<option object>, …],
     "verbs": [<command >, …],
-    "documentation": ["doc1", …]
+    "documentation": ["doc1", …],
+    `package`: "myproject"
 }
 ```
 
@@ -100,6 +101,9 @@ that should not be larger than 16.
 
 `documentation` is an array of string values describing URIs referencing
 documentation for this command, see `man:uri(7)`for a description of valid URIs.
+
+`package` is a string describing what this command belongs to. This may the
+package that installed it or the project that produced it.
 
 ### Argspec objects
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -64,7 +64,7 @@ The command object is recursively defined as follows.
     "version": "myversion",
     "features": ["feature1", "feature2", …]
     "abstract": ["paragraph1", "paragraph2", …],
-    "footer": ["paragraph1", "paragraph2", …],
+    "postscript": ["paragraph1", "paragraph2", …],
     "options": [<option object>, …],
     "verbs" : [<command >, …]
 }
@@ -83,7 +83,7 @@ UAPI.10 Version Format Specification-compatible.
 `features` defines an array of strings, that can be empty, that define
 builtin-in features of this program.
 
-`abstract` and `footer` define arrays of strings that are shown respectively
+`abstract` and `postscript` define arrays of strings that are shown respectively
 before and after the description of options in the program's help output. These
 arrays may be empty.
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -61,6 +61,7 @@ The command object is recursively defined as follows.
 ```json
 {
     "names" : ["name1", "name2", …],
+    "argspec": ["value1", "value2", …]
     "version": "myversion",
     "features": ["feature1", "feature2", …]
     "abstract": ["paragraph1", "paragraph2", …],
@@ -77,6 +78,9 @@ array when missing.
 of that array is the primary name of that command. Further names can be added as
 aliases, e.g. for backwards-compatibility.
 
+`argspec` is an array of argspec objects describing the arguments to the command
+after the name of the command, i.e. `argv` excluding the element with index 0.
+
 `version` defines the version string of the program. This string should be a
 UAPI.10 Version Format Specification-compatible.
 
@@ -92,6 +96,43 @@ arrays may be empty. Each element of the array represents a paragraph of text.
 `verbs` are command objects that describe a program's verbs, also called
 subcommands. Verbs may recursively define further verbs up to a maximum depth
 that should not be larger than 16.
+
+### Argspec objects
+
+Argspec objects describe the arguments passed to the program, specifically in
+what order arguments, verbs and options are passed to the program.
+
+```json
+{
+    "name": "myname",
+    "metavar": "METAVAR",
+    "type": "argument",
+    "optional": true,
+    "default": null
+}
+```
+
+All keys except `name` and `type` are required and are treated as empty string,
+false or null if missing.
+
+`name` is a non-empty string describing the internal name of a value.
+
+`metavar` is a string describing the name that is shown in the argspec of the
+command. If it is empty, it defaults to the value of `name`.
+
+`type` is a string describing what the argument is. It is either of
+- `argument`, a direct argument to the command,
+- `option`, which is an option from the `options` array of a command, or
+- `verb`, which is a verb from the `verbs` array of a command.
+
+`optional` is a boolean describing whether an argspec object of type `argument`
+or `verb` must be passed or not. `optional` is ignored for argspec objects of
+type `option`.
+
+`default` is a string or null describing what the default value for this
+argument is. `null`, the default, means there is no default. `default` is
+ignored for argspec objects of type `option`. For argspec objects of type `verb`
+a non-`null` value must be a `name` of a commands object in the verbs array.
 
 ### Option objects
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -73,7 +73,8 @@ The command object is recursively defined as follows.
     "options": [<option object>, …],
     "verbs": [<command >, …],
     "documentation": ["doc1", …],
-    "project": "myproject"
+    "project": "myproject",
+    "isDeprecated": false
 }
 ```
 
@@ -110,6 +111,8 @@ widely supported in modern terminals.
 
 `project` is a string describing what this command belongs to. This may the
 package that installed it or the project that produced it.
+
+`isDeprecated` is a boolean describing whether the command has been
 
 ### Argspec objects
 
@@ -162,7 +165,8 @@ explicitly made.
     "help": "Show this help",
     "group": [""],
     "values": [<value object>],
-    "section": "option block 1"
+    "section": "option block 1",
+    "isDeprecated": false
 }
 ```
 
@@ -194,6 +198,8 @@ take. Value objects are described in a section below.
 
 `section` is a string to visually group options into separate lists.
 
+`isDeprecated` is a boolean describing whether this option has been deprecated.
+
 ### Value objects
 
 Value objects describe a value that is passed as an argument, either to a
@@ -203,7 +209,8 @@ Value objects describe a value that is passed as an argument, either to a
 {
     "name": "myname",
     "help": "help text",
-    "default": true
+    "default": true,
+    "isDeprecated": false
 }
 ```
 
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -1,0 +1,160 @@
+---
+title: UAPI.X CLI-Introspection Specification
+category: Booting
+layout: default
+version: 0.0
+SPDX-License-Identifier: CC-BY-4.0
+weight: 1
+aliases:
+- /UAPI.X
+- /X
+---
+
+# UAPI.X The CLI-Introspection Specification
+
+| Version | Changes |
+|---------|---------|
+| 0.0     | WIP     |
+
+This document defines a JSON schema defining a machine-readable equivalent of
+what programs commonly output when they are called with the `--help` or an
+equivalent option.
+
+This output can be used, among other things, to check whether a program has a
+given option or to automatically generate command-line completion for different
+shells.
+
+## Target Audience
+
+The target audience for this specification are:
+
+* People writing runnable programs that have command-line interfaces.
+
+## The Schema
+
+The schema is defines as
+
+```json
+{
+    "mediaType" : "application/vnd.io.systemd.cli-introspection-0",
+    "commands": [<command object>, …]
+}
+```
+
+where `mediaType` is the fixed string
+`application/vnd.io.systemd.cli-introspection-0` and `commands` is a non-empty
+JSON array of *command objects*.
+
+Further additions to this specification will only be additive and non-backwards
+compatible changes may only be introduced by defining a new `mediaType`.
+
+The `commands` array defines the actual CLI and a single binary can define
+arbitrarily many CLIs, but must define at least one. Defining multiple CLIs is
+relevant for multicall binaries that behave differently depending on with what
+name they are called, e.g `argv[0]` or in the context of scripting languages
+under what name they are imported
+
+### Command ojects
+
+The command object is recursively defined as follows.
+
+```json
+{
+    "names" : ["name1", "name2", …],
+    "version": "myversion",
+    "features": ["feature1", "feature2", …]
+    "abstract": ["paragraph1", "paragraph2", …],
+    "footer": ["paragraph1", "paragraph2", …],
+    "options": [<option object>, …],
+    "verbs" : [<command >, …]
+}
+```
+
+All keys except `names` are optional and are treated as empty string or empty
+array when missing.
+
+`names` is a non-empty array of string names for that command. The first element
+of that array is the primary name of that command. Further names can be added as
+aliases, e.g. for backwards-compatibility.
+
+`version` defines the version string of the program. This string should be a
+UAPI.10 Version Format Specification-compatible.
+
+`features` defines an array of strings, that can be empty, that define
+builtin-in features of this program.
+
+`abstract` and `footer` define arrays of strings that are shown respectively
+before and after the description of options in the program's help output. These
+arrays may be empty.
+
+`options` define a program's arguments and options and are described in a section below.
+
+`verbs` are command objects that describe a program's verbs, also called
+subcommands. Verbs may recursively define further verbs up to a maximum depth
+that should not be larger than 16.
+
+### Option objects
+
+Option objects arguments define both (positional)arguments as well as options
+(commonly prefixed with `-` or `--`) of a program. Both arguments and options
+will be referred to as options in this section, when a distinction is not
+explicitly made.
+
+```json
+{
+    "names": ["-h","--help"],
+    "argument": "no_argument",
+    "help": "Show this help",
+    "group": [""],
+    "conflicts": [],
+    "values": [<value object>]
+}
+```
+
+All keys except `names` and `argument` are optional and are treated as empty
+string or empty array when missing.
+
+`names` is a non-empty array of strings defining the name of an option. Names
+starting with dashes define options and names not starting with dashes define
+(positional) arguments. An option object may define only options or arguments,
+but not both for the same object, i.e. an option object may not have names both
+with and without dashes. Options prefixed with a single dash (`-`) are called
+short options and options prefixed with two dashes (`--`) are called
+long-options. Short options are usually followed by a single chracters, whereas
+long options can be a longer string. Short otions may be followed by multiple
+characters, but this is discouraged.
+
+`argument` defines the argument type TODO
+
+`help` is a string that defines the help text of the option.
+
+`group` is an array of string that defines sections in which this option should
+be shown. This is only for display-purposes.
+
+`conflicts` defines an array of strings referring to the names of other options
+this option may not be called together with.
+
+`values` is an array of *value objects* describing the values this option may
+take. Value objects are described in a section below.
+
+### Value objects
+
+Value objects describe a value that is passed as an argument, either to a
+(positional) argument or an option.
+
+```json
+{
+    "name": "myname",
+    "help": "help text",
+    "default": True
+}
+
+All keys except `name` are optional and are treated as empty string or false
+when missing.
+
+`name` is a string describing the name of a value.
+
+`help` is a string describing the help text that should be shown for an option.
+
+`default` is a boolean describing whether this value is the default value for
+the option this is a value for.
--- a/specs/cli_introspection.md
+++ b/specs/cli_introspection.md
@@ -1,13 +1,13 @@
 ---
 title: UAPI.X CLI-Introspection Specification
-category: Booting
+category: Interfaces
 layout: default
 version: 0.0
 SPDX-License-Identifier: CC-BY-4.0
 weight: 1
 aliases:
-- /UAPI.X
-- /X
+- /UAPI.18
+- /18
 ---
 
 # UAPI.X The CLI-Introspection Specification
@@ -28,15 +28,18 @@ shells.
 
 The target audience for this specification are:
 
-* People writing runnable programs that have command-line interfaces.
+* People writing runnable programs that have command-line interfaces,
+* authors of tab-completion helpers, and
+* people interested in automatic testing of documentation and self-documenting
+  interfaces.
 
 ## The Schema
 
-The schema is defines as
+The schema is defined as
 
 ```json
 {
-    "mediaType" : "application/vnd.uapi-group.cli-introspection-0",
+    "mediaType" : "application/vnd.uapi-group.cli-introspection",
     "commands": [<command object>, …]
 }
 ```
@@ -45,14 +48,15 @@ where `mediaType` is the fixed string
 `application/vnd.uapi-group.cli-introspection-0` and `commands` is a non-empty
 JSON array of *command objects*.
 
-Further additions to this specification will only be additive and non-backwards
+Further additions to this specification will only be additive and non-backward
 compatible changes may only be introduced by defining a new `mediaType`.
 
 The `commands` array defines the actual CLI and a single binary can define
 arbitrarily many CLIs, but must define at least one. Defining multiple CLIs is
 relevant for multicall binaries that behave differently depending on with what
-name they are called, e.g `argv[0]` or in the context of scripting languages
-under what name they are imported
+name they are called, e.g the basename of `argv[0]`,
+`program_invocation_short_name` in the GNU system context, or in the context of
+scripting languages the name under which they are imported
 
 ### Command ojects
 
@@ -78,29 +82,31 @@ array when missing.
 
 `names` is a non-empty array of string names for that command. The first element
 of that array is the primary name of that command. Further names can be added as
-aliases, e.g. for backwards-compatibility.
+aliases, e.g. for backward compatibility.
 
 `argspec` is an array of argspec objects describing the arguments to the command
-after the name of the command, i.e. `argv` excluding the element with index 0.
+after the name of the command, i.e. C language `argv` excluding the element with index 0.
 
-`version` defines the version string of the program. This string should be a
-UAPI.10 Version Format Specification-compatible.
+`version` defines the version string of the program. This string should be string compatible
+with UAPI.10 Version Format Specification-compatible.
 
-`features` defines an array of strings, that can be empty, that define
-builtin-in features of this program.
+`features` defines an array of strings that define builtin-in features of this
+program.
 
 `abstract` and `postscript` define arrays of strings that are shown respectively
-before and after the description of options in the program's help output. These
-arrays may be empty. Each element of the array represents a paragraph of text.
+before and after the description of options in the program's help output. Each
+element of the array represents a paragraph of text.
 
-`options` define a program's arguments and options and are described in a section below.
+`options` define a CLI's positional arguments and options and are described in a section below.
 
 `verbs` are command objects that describe a program's verbs, also called
 subcommands. Verbs may recursively define further verbs up to a maximum depth
 that should not be larger than 16.
 
 `documentation` is an array of string values describing URIs referencing
-documentation for this command, see `man:uri(7)`for a description of valid URIs.
+documentation for this command, see `man:uri(7)`for a description of valid
+URIs. Preferably the URIs should be URLs starting with `https://` as these are
+widely supported in modern terminals.
 
 `project` is a string describing what this command belongs to. This may the
 package that installed it or the project that produced it.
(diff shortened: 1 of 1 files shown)


specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+* People writing runnable programs that have command-line interfaces.
+
+## The Schema
+
+The schema is defines as
+
+```json
+{
+    "mediaType" : "application/vnd.uapi-group.cli-introspection-0",
```
behrmann: > hm, the …-0 suffix, do we really want that? is ther prior art?

So that is mostly because I'm assuming that this will need to go through iterations. Following the discussion in #213 I agree that having a version field is not worthwhile and if this needs a breaking change it will get a new `mediaType`. The `-0` is for symmetry` once that get's bumped, so that the first one also has a number.

> and we should change this to some uapi string, i.e. drop the systemd part

Changed to `vnd.uapi-group`.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+The command object is recursively defined as follows.
+
+```json
+{
+    "names" : ["name1", "name2", …],
+    "argspec": ["value1", "value2", …]
+    "version": "myversion",
+    "features": ["feature1", "feature2", …]
+    "abstract": ["paragraph1", "paragraph2", …],
+    "postscript": ["paragraph1", "paragraph2", …],
```
behrmann: > i think we should find a better name for "footer", as that indicates location where to print that, not the category of the text it contains.

Changed it to `postscript`, but I'm open to other things.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+long-options. Short options are usually followed by a single chracters, whereas
+long options can be a longer string. Short otions may be followed by multiple
+characters, but this is discouraged.
+
+`argument` defines the argument type, which is one of the strings:
+-`no_argument`,
+-`required_argument`, or
+-`optional_argument`.
+
+`help` is a string that defines the help text of the option.
```
behrmann, line 248: > hmm, you call this "help" here, and "abstract for the verbs. we should figure this out, and stick to the same concept

I've still got adding help stashed, because the command objects describe both the top-level program and verbs (so this can cover subcommands in git's style, something systemd doesn't really do). In this case a short help string for a verb makes sense and abstract and postscript don't.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    "names" : ["name1", "name2", …],
+    "argspec": ["value1", "value2", …]
+    "version": "myversion",
+    "features": ["feature1", "feature2", …]
+    "abstract": ["paragraph1", "paragraph2", …],
+    "postscript": ["paragraph1", "paragraph2", …],
+    "options": [<option object>, …],
+    "verbs": [<command >, …],
+    "documentation": ["doc1", …],
+    "package": "myproject"
```
behrmann: As mentioned in https://github.com/systemd/systemd/pull/43430 I'd personally go with `.project` here.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+arrays may be empty. Each element of the array represents a paragraph of text.
+
+`options` define a program's arguments and options and are described in a section below.
+
+`verbs` are command objects that describe a program's verbs, also called
+subcommands. Verbs may recursively define further verbs up to a maximum depth
+that should not be larger than 16.
+
+`documentation` is an array of string values describing URIs referencing
+documentation for this command, see `man:uri(7)`for a description of valid URIs.
```
behrmann: As mentioned in https://github.com/systemd/systemd/pull/43152 I really like the style `man:foo.sec` used there, but this currently clashes with the definitions in `man:uri(7)`. I think this works because man reads the dot as the file suffix. Changing this might be difficult, since it would need an IETF RFC?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
+---
+title: UAPI.X CLI-Introspection Specification
+category: Booting
```
keszybz: `category: Interfaces`

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+# UAPI.X The CLI-Introspection Specification
+
+| Version | Changes |
+|---------|---------|
+| 0.0     | WIP     |
+
+This document defines a JSON schema defining a machine-readable equivalent of
+what programs commonly output when they are called with the `--help` or an
+equivalent option.
```
keszybz, line 21: ~~or an equivalent option~~

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+This output can be used, among other things, to check whether a program has a
+given option or to automatically generate command-line completion for different
+shells.
+
+## Target Audience
+
+The target audience for this specification are:
+
+* People writing runnable programs that have command-line interfaces.
```
keszybz: Also: authors of tab-completion helpers, people interested in automatically generating or checking documentation.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+## The Schema
+
+The schema is defines as
```
keszybz: "defined"

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    "mediaType" : "application/vnd.uapi-group.cli-introspection-0",
+    "commands": [<command object>, …]
+}
+```
+
+where `mediaType` is the fixed string
+`application/vnd.uapi-group.cli-introspection-0` and `commands` is a non-empty
+JSON array of *command objects*.
+
+Further additions to this specification will only be additive and non-backwards
```
keszybz: "backward compatible" (no "s").

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+compatible changes may only be introduced by defining a new `mediaType`.
+
+The `commands` array defines the actual CLI and a single binary can define
+arbitrarily many CLIs, but must define at least one. Defining multiple CLIs is
+relevant for multicall binaries that behave differently depending on with what
+name they are called, e.g `argv[0]` or in the context of scripting languages
+under what name they are imported
```
keszybz: "the name under which they are imported"

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    "postscript": ["paragraph1", "paragraph2", …],
+    "options": [<option object>, …],
+    "verbs": [<command >, …],
+    "documentation": ["doc1", …],
+    "project": "myproject"
+}
+```
+
+All keys except `names` are optional and are treated as empty string or empty
+array when missing.
```
keszybz: Just "treated as empty." would be enough.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`names` is a non-empty array of string names for that command. The first element
+of that array is the primary name of that command. Further names can be added as
+aliases, e.g. for backwards-compatibility.
```
keszybz: "backward compatibility"

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`argspec` is an array of argspec objects describing the arguments to the command
+after the name of the command, i.e. `argv` excluding the element with index 0.
```
keszybz: maybe mention "C language `argv[]`" here. Might be unclear for non-C programmers.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`version` defines the version string of the program. This string should be a
+UAPI.10 Version Format Specification-compatible.
```
keszybz: "Should be compatible with …"

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`features` defines an array of strings, that can be empty, that define
```
keszybz: ~~that can be empty~~ (because it's not clear if the array or strings can be empty, and if it's the array, this was already mentioned at the top).

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+builtin-in features of this program.
+
+`abstract` and `postscript` define arrays of strings that are shown respectively
+before and after the description of options in the program's help output. These
+arrays may be empty. Each element of the array represents a paragraph of text.
```
keszybz: Again, no need to mention optional emptiness.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`options` define a program's arguments and options and are described in a section below.
```
keszybz: "program" → "CLI"? "positional arguments and options"?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+`documentation` is an array of string values describing URIs referencing
+documentation for this command, see `man:uri(7)`for a description of valid URIs.
+
+`project` is a string describing what this command belongs to. This may the
+package that installed it or the project that produced it.
+
+### Argspec objects
+
+Argspec objects describe the arguments passed to the program, specifically in
+what order arguments, verbs and options are passed to the program.
```
keszybz: "verbs, options, and positional arguments"? Oxford comma and also positional arguments are after verbs.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+```json
+{
+    "name": "myname",
+    "metavar": "METAVAR",
```
keszybz: We called this "metavar" I think because of Python's argspec. But that is not a great name. Clap uses "value_name" (https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_name), which seems much better.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+command. If it is empty, it defaults to the value of `name`.
+
+`type` is a string describing what the argument is. It is either of
+- `argument`, a direct argument to the command,
+- `option`, which is an option from the `options` array of a command, or
+- `verb`, which is a verb from the `verbs` array of a command.
+
+`optional` is a boolean describing whether an argspec object of type `argument`
+or `verb` must be passed or not. `optional` is ignored for argspec objects of
+type `option`.
```
keszybz: I'd drop that second sentence. I don't think we need to specify this here, and some projects do have "required options".

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`default` is a string or null describing what the default value for this
+argument is. `null`, the default, means there is no default. `default` is
+ignored for argspec objects of type `option`. For argspec objects of type `verb`
```
keszybz: Why ignored?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+a non-`null` value must be a `name` of a commands object in the verbs array.
+
+### Option objects
+
+Option objects arguments define both (positional)arguments as well as options
```
keszybz: space missing

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    "argument": "no_argument",
+    "help": "Show this help",
+    "group": [""],
+    "conflicts": [],
+    "values": [<value object>],
+    "section": "option block 1"
+}
+```
+
+All keys except `names` and `argument` are optional and are treated as empty
```
keszybz: Also here, "treated as empty when missing" is enough.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+string or empty array when missing.
+
+`names` is a non-empty array of strings defining the name of an option. Names
+starting with dashes define options and names not starting with dashes define
+(positional) arguments. An option object may define only options or arguments,
```
keszybz: Drop "()" ?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+but not both for the same object, i.e. an option object may not have names both
+with and without dashes. Options prefixed with a single dash (`-`) are called
+short options and options prefixed with two dashes (`--`) are called
+long-options. Short options are usually followed by a single chracters, whereas
```
keszybz: "long options"

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+    "name": "myname",
+    "help": "help text",
+    "default": true
+}
+```
+
+All keys except `name` are optional and are treated as empty string or false
+when missing.
+
+`name` is a string describing the name of a value.
```
keszybz: "value_name" for consistency?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`help` is a string describing the help text that should be shown for an option.
```
keszybz: "for the value"?

keszybz (changes requested):
What about using semantic breaks? I think it'd also make making review comment easier.

There should be a complete example, or maybe even more than one.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
+---
+title: UAPI.X CLI-Introspection Specification
+category: Booting
+layout: default
+version: 0.0
+SPDX-License-Identifier: CC-BY-4.0
+weight: 1
+aliases:
+- /UAPI.X
```
poettering: please assign uapi 18 here

specs/cli_introspection.md
poettering: so i used to suggest that dbus interfaces are versioned with "1" in their initial version too back in the day. But nowadays i think it's a mistake, because in effect none of this dbus interfaces ever got a breaking update. people who are smart enough to consider versioning tend to be also smart enough to avoid breaking changes somehow. at least i never encountered a dbus interface where people suffixed a version where the version was 2 or above...

hence, i'd just not do it for the first version, but only do it once we do the first compat break.

or in other words, instead of doing a series of "foo-0, foo-1, foo-2" just do "foo, foo-2, foo-3".

specs/cli_introspection.md
poettering: maybe reference "program_invocation_short_name" here, as the suggested identifier string for the CLI, at least on gnu systems

specs/cli_introspection.md
poettering: in the brief version above you suggest the array contains strings, here you write "objects". what is it?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+arrays may be empty. Each element of the array represents a paragraph of text.
+
+`options` define a program's arguments and options and are described in a section below.
+
+`verbs` are command objects that describe a program's verbs, also called
+subcommands. Verbs may recursively define further verbs up to a maximum depth
+that should not be larger than 16.
+
+`documentation` is an array of string values describing URIs referencing
+documentation for this command, see `man:uri(7)`for a description of valid URIs.
```
poettering: i think we really shoudl recommend man: and https:// urls, and discourage everything else.

man: urls might not be standard, but rlevant terminals all support them.

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+All keys except `name` and `type` are required and are treated as empty string,
+false or null if missing.
+
+`name` is a non-empty string describing the internal name of a value.
+
+`metavar` is a string describing the name that is shown in the argspec of the
+command. If it is empty, it defaults to the value of `name`.
+
+`type` is a string describing what the argument is. It is either of
+- `argument`, a direct argument to the command,
```
poettering: is this supposed to be what gnu getopt() docs call "positional arguments"?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+- `option`, which is an option from the `options` array of a command, or
```
poettering: these are switches, right?

specs/cli_introspection.md
poettering: very confusing that option and optional both exist, and do unrelated things?

maybe rename "option" to "switch"?

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+-`required_argument`, or
+-`optional_argument`.
+
+`help` is a string that defines the help text of the option.
+
+`group` is an array of string that defines sections in which this option should
+be shown. This is only for display-purposes.
+
+`conflicts` defines an array of strings referring to the names of other options
+this option may not be called together with.
```
poettering: i'd probably leave this out of the initial version of the spec (i doubt we'd fill this in in systemd anytime soon, right?), unless there's some other project that very clearly wants to include this

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`values` is an array of *value objects* describing the values this option may
+take. Value objects are described in a section below.
```
poettering: maybe make clear that absence or null of this means that the syntax/dictionary of the values is left open

poettering:
verb objects are not defined yet, but i guess you know that

specs/cli_introspection.md
keszybz: > people who are smart enough to consider versioning tend to be also smart enough to avoid breaking changes somehow

:heart:

specs/cli_introspection.md
```diff
@@ -0,0 +1,216 @@
...
+
+`abstract` and `postscript` define arrays of strings that are shown respectively
+before and after the description of options in the program's help output. These
+arrays may be empty. Each element of the array represents a paragraph of text.
+
+`options` define a program's arguments and options and are described in a section below.
+
+`verbs` are command objects that describe a program's verbs, also called
+subcommands. Verbs may recursively define further verbs up to a maximum depth
+that should not be larger than 16.
```
keszybz: Should verbs have their own `features` and `version` and `project` fields? Maybe say that those are not allowed except at the top level?

keszybz:
> verb objects are not defined yet, but i guess you know that

They are defined as recursive commands.

poettering:
> They are defined as recursive commands.

ah, i see that now. not sure that's ideal though, because some of the props of commands don't really apply to verbs, do they?

dunno, i am a bit unsure about thsi.

keszybz:
During implementation I realized that there should be a way to mark both verbs and options as deprecated. Maybe even commands.

EDIT: I'm now using optional 'isDeprecated: true'.

specs/cli_introspection.md
behrmann: the reason for that formulation is that `-h` is also common and in golang world, much to my dismay, `-help` seems to be quite common as well. I've also seen the quite disgusting `-?` there once or twice.

keszybz:
Another RFE: indicate that some part of the description is missing. For example, if options are not described, or helps are excluded to save space, etc. And in the future, if we include a description of how to generate values for an option, that is missing for an option. For example, because the completions should be generated dynamically.

behrmann:
I pushed some changes addressing the above remarks, but after implementing them I mostly rewrote this to get rid of the argspec array. 

The argspec array was necessary because options and verbs were treated separately and therefore no order between options, positional arguments and verbs was established. The split was awkward, because of the need to consult multiple arrays in the object referencing each other. Combining all arguments into a single array established a natural ordering and makes the information more local. This also gets red of some of the option/optional confusions. Unfortunately it meant that the type from the argspec objections needed to be pushed down, because now objects of different shapes live in the same array. For consistency everything got a type field.

I also added a nomenclature section and added a worked example.

The latest commit address @keszybz most recent RFE.

specs/cli_introspection.md
behrmann: It does irk my sense for symmetry a bit, but that's reasonable. :)

specs/cli_introspection.md
behrmann: The worked example has got both, I'll leave this open to add the encouragement for `man:`

specs/cli_introspection.md
behrmann: this has been dropped for now.

specs/cli_introspection.md
behrmann: I added a formulation regarding this. PTAL whether the intent "this is some arbitrary blob we don't say anything about" comes across.

poettering:
given that we are going to release systemd v262 soon, where this is supported, is it time to undraft and stabilize this?

behrmann:
> given that we are going to release systemd v262 soon, where this is supported, is it time to undraft and stabilize this?

I think the current version in systemd is not quite what's in this spec at the moment (@keszybz will know best) and there's one or two more things I've not pushed yet, because `$DAYJOB` kept me busy these past two weeks and I've not had time to work on this since BtO. 

The introspection stuff is has a note

> This feature
          is experimental and the JSON schema may be subject to change in later
          release until a future stabilization announcement.

in the NEWS for 262, so this can still change, but I think releasing the spec is more something for after this stabilizes and becomes useful in systemd.

julian-klode:
I think this maps reasonably well to my half-design for APT's declarative command-line parser, but it would be useful to  be a bit more explicit about help messages. 

In my mental model (not fully implemented), I define command help as usage, summary and long description - the pattern we go for being
- Main help shows the commands and their summary
- Command-specific help is `Usage: apt <command> ...` (possibly multiple lines) followed by the long description.

I can map this to the spec to map usage to `arguments`, `help` for the short description, and `abstract` the long one.

One of the things that we likely can't address well here is multi-mode commands, notably we have:

```
Usage: apt [options] why <package>
       apt [options] why-not <package>

Explain why a package is or is not installed. The why command
shows why an installed package is present; why-not shows why a
package cannot be installed. Takes a single package name.
```

And similarly say `install` and `remove` - it often makes sense to document them together if they have the same flags but inverse behavior. We'd instead need to repeat ourselves.

One thing that could be more strictly defined to be useful is the `help` field - I'm thinking about `fish` completions which show a short description next to each item. The field is a bit loosely defined and reads like I could add a second-line, which sometimes is useful for `--help`, say:

```
--foo   Do bar. 
            This is not the best thing in the world, consider using --bar instead.  # would not want this in fish
```

Finally, option groups are often shared for us by up to 10 verbs, having to repeat ourselves is a bit of an overhead but it's likely not the end of the world.

specs/cli_introspection.md
keszybz: ack

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+
+```json
+{
+  "mediaType" : "application/vnd.uapi-group.cli-introspection",
+  "commands": [<command object>, …]
+}
+```
+
+where `mediaType` is the fixed string
+`application/vnd.uapi-group.cli-introspection-0` and `commands` is a non-empty
```
keszybz, line 49: Drop "-0" here.

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+`documentation` is an array of string values describing URIs referencing
+documentation for this command,
+see `man:uri(7)`for a description of valid URIs.
+Preferably the URIs should be URLs starting with `https://` as these are widely
+supported in modern terminals.
+
+`project` is a string describing what this command belongs to.
+This may the package that installed it or the project that produced it.
+
+`isDeprecated` is a boolean describing whether the command has been
```
keszybz, line 156: … something?

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+### Argument objects
+
+Argument objects describe positional arguments that are not verbs.
+
+```json
+{
+  "type": "argument"
+  "name": "filename",
+  "value_name": "FILE",
+  "help": "Show this help",
```
keszybz, line 167: This should be something like "filename to operate on", since this is an arg.

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+-`optional_argument`.
+An argument object whose `argument` value is `required_argument` must be
+passed an argument,
+while an argument of whose `argument` value is `optional_argument` may be omitted.
+
+`help` is a string that defines the help text of the argument.
+
+`sections` is an array of string that defines sections in which this option should
+be shown.
+This is only for display-purposes.
```
keszybz, line 193: I think "display purposes" is more usual.

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+
+`names` is a non-empty array of strings defining the name of an option.
+Names starting with dashes define options and names not starting with dashes
+define positional arguments.
+An option object may define only options or arguments,
+but not both for the same object,
+i.e. an option object may not have names both with and without dashes.
+Options prefixed with a single dash (`-`) are called short options and options
+prefixed with two dashes (`--`) are called long options.
+Short options are usually followed by a single character,
```
keszybz, line 233: "Short options are usually just a single character after the dash"

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+whereas long options can be a longer string.
+Short options may be followed by multiple characters,
```
keszybz, line 235: I would rephase this. A single-dash option with a long name is not a short option, but a long option with just one dash ;(

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+        "+LZ4",
+        "+XZ",
+        "+ZLIB",
+        "+ZSTD",
+        "+BPF",
+        "FRAMEWORK",
+        "+BTF",
+        "+XKBCOMMON",
+        "+UTMP",
+        "+LIBARCHIVE"
```
keszybz, line 405: In the spec we could trim this to maybe 5 lines. We don't need all the systemd internals here.

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+  interfaces.
+
+## The Schema
+
+The schema is defined as
+
+```json
+{
+  "mediaType" : "application/vnd.uapi-group.cli-introspection",
+  "commands": [<command object>, …]
```
poettering, line 44: Hmm, should this be a dictionary btw? Witht the command name as key and the parameters object as value?

poettering: I guess the answer is "no" because alias names shall be supported? Ignore me then

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+`abstract` and `postscript` define arrays of strings that are shown respectively
+before and after the description of options in the program's help output.
+Each element of the array represents a paragraph of text.
+Both are meant for standalone help output, e.g. for the top-level help output of
+a program or specific help output of a verb in cases where they have their own
+help output.
+
+`help` defines the help text of the command. This is useful for single-line
+usage information as well as for short descriptions of verbs in the list of
+options.
```
poettering, line 145: I think we need to make clearer when to output abstract/postscript and when to output help.

As i understand you pick either abstract+postscript OR help, but not both at the same time, right?

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+The depth of this object,
+since command objects in this array may themselves define command objects in
+their `arguments`,
+may not exceed 15.
+
+`documentation` is an array of string values describing URIs referencing
+documentation for this command,
+see `man:uri(7)`for a description of valid URIs.
+Preferably the URIs should be URLs starting with `https://` as these are widely
+supported in modern terminals.
```
poettering: What about man: urls? Shouldnt those be preferable?

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+
+### Argument objects
+
+Argument objects describe positional arguments that are not verbs.
+
+```json
+{
+  "type": "argument"
+  "name": "filename",
+  "value_name": "FILE",
```
poettering: The spec otherwise uses lowecase camelCase hence here: "valueName" rather than "value_name"

specs/cli_introspection.md
poettering: Should say if unspecified defaults to false

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+be shown.
+This is only for display-purposes.
+
+`values` is an array of *value objects* describing the values this option may
+take.
+An empty array describes an argument that is a single arbitrary word with no
+further documented semantic.
+Value objects are described in a section below.
+
+`isDeprecated` is a boolean describing whether this option has been deprecated.
```
poettering, line 247: As above

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+  "help": "short help"
+  "arguments": [<option|argument|command>, …],
+  "documentation": ["doc1", …],
+  "project": "myproject",
+  "isDeprecated": false
+}
+```
+
+All keys except `type` and `names` are optional and are treated as empty string
+or empty array when missing.
```
poettering: Well isDeprecated is neither string nor array. I think for non-string/non-array we should document the defaults separately.

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+describing a positional argument.
+
+`names` is a non-empty string defining the name of the argument.
+
+`argument` defines the argument type, which is one of the strings:
+-`required_argument`, or
+-`optional_argument`.
+An argument object whose `argument` value is `required_argument` must be
+passed an argument,
+while an argument of whose `argument` value is `optional_argument` may be omitted.
```
poettering: Should say what mode is default if unspecified

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+
+`features` defines an array of strings that define builtin-in features of this
+program.
+
+`abstract` and `postscript` define arrays of strings that are shown respectively
+before and after the description of options in the program's help output.
+Each element of the array represents a paragraph of text.
+Both are meant for standalone help output, e.g. for the top-level help output of
+a program or specific help output of a verb in cases where they have their own
+help output.
```
poettering, line 141: Maybe we should emphasize that the string should not be linebroken, and the consumer is supposed to line break it for display purposes.

specs/cli_introspection.md
behrmann: Yes, that's true and I greatly dislike multi-letter "short" options, but it it makes sense to call them short options here in the context of the previous sentences, since two sentences prior short options are defined by starting with a single dash. I inserted another half-sentence to clarify this.

specs/cli_introspection.md
behrmann: I added a section discussing the difference, I think there are scenarios were setting all of them or both `help` and `abstract` makes sense.

specs/cli_introspection.md
behrmann: Ah, I had misunderstood your comment above, I'll address that.

My thinking was that URLs are nicely clickable in modern terminal emulators, so I thought they should be preferred over `man:`

specs/cli_introspection.md
behrmann: Changed, though `value_name` predates the first camelCase keyword that I added after @keszybz added that in systemd, I'm more of a snake_case kind of person. ;)

specs/cli_introspection.md
behrmann: I added that in the two missing places, since for all objects the non-optional attributes are described above and all optional ones should be considered empty/falsey, i.e. empty strings, empty arrays, empty objects, false, …

The listing was a bit more explicit, but I shortened it to empty due to some earlier comment, so the false got lost.

specs/cli_introspection.md
behrmann: discussed above

specs/cli_introspection.md
behrmann: `argument` is non-optional, so it must be specified.

specs/cli_introspection.md
behrmann: Added.

behrmann:
I added two commits addressing the comments and I added a section discussing the respective use of `help` vs `abstract`/`postscript`. The idea is that having a command `foo` with help

```
foo [opt] ARG

Here I explain what foo is about.

This can be longer.

Verbs:
    bar    bar does some bar things

Thanks for reading to the end
```

`["Here I explain what foo is about.", "This can be longer."]` would be `foo`'s `abstract`, `["Thanks for reading to the end"]` would be `foo`'s `postscript` and "bar does some bar things" would be `bar`'s `help`.

It can make sense for `bar` to define `abstract` and `postscript`, though, if `foo bar --help` has a different output than `foo --help`, which is common for git subcommands (though they only show options and don't have help text) or subcommands with Python's argparse.

It can also make sense for `foo` to define `help`, if e.g. calling `foo` without arguments show a single usage line.

@julian-klode I think my description maps somewhat onto what you are describing for help output, does this make sense to you?

specs/cli_introspection.md
poettering: man: works fine too, at least on gnome. and usually local stuff might be nicer, dunno

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+  "name": "filename",
+  "valueName": "FILE",
+  "help": "filename to operate on",
+  "sections": ["Arguments"],
+  "values": [<value object>],
+  "isDeprecated": false
+}
+```
+
+All keys except `type`, `name` and `argument` are optional and are treated as
```
poettering: `argument` is not in the json structure above but is supposedly required.

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+describing a positional argument.
+
+`names` is a non-empty string defining the name of the argument.
+
+`argument` defines the argument type, which is one of the strings:
+-`required_argument`, or
+-`optional_argument`.
+An argument object whose `argument` value is `required_argument` must be
+passed an argument,
+while an argument of whose `argument` value is `optional_argument` may be omitted.
```
poettering: actually, i think we should call the values here `required` and `optional` instead of `required_argument` and `optional_argument`. it's needlessly repetitive and redundant to repeat the field name like this in the enum values.

poettering: (i mean, otherwise if you wanted to be systematic, you'd have to call the values that you can set in `type`  to `argument_type`, `option_type` and so on – rather than `argument` and `option`– and that'd be weird)

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+
+`argument` defines the argument type, which is one of the strings:
+-`no_argument`,
+-`required_argument`, or
+-`optional_argument`.
+An argument object whose `argument` value is `no_argument` cannot be passed an
+argument,
+an argument object whose `argument` value is `required_argument` must be
+passed an argument,
+and an argument of whose `argument` value is `optional_argument` may be omitted.
```
poettering: as above `no`, `required`, `optional`

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+}
+```
+
+All keys except `type`, `name` and `argument` are optional and are treated as
+empty or false when missing.
+
+`type` is the fixed string `argument` and signals that this is an argument object,
+describing a positional argument.
+
+`names` is a non-empty string defining the name of the argument.
```
poettering: hmm, so positional argument are identified by their position, and hence giving them a name is surprising to some level. I think it *does* make sense to give them a name, simply for internal identification purposes, but i think this needs to be discussed here, in particular as for positional arguments we never should show this, should we? i.e. the `name` field of the option arguments is very visible, but this here is not.

poettering: hmm, maybe we should get rid of the `names` field for positional arguments, and instead introduce an `id` field or so, which is used as internal identifier for things, and should be available on all argument types. This would not only be useful here for the positional arguments but also for options, since for it might be nice to an id for "-h"/"--help" that is not prefixed with "-", and is one string instead of two.

this id should probably be optional though.

specs/cli_introspection.md
```diff
@@ -0,0 +1,637 @@
...
+usage information as well as for short descriptions of verbs in the list of
+options.
+
+`arguments` defines a CLI's positional and optional arguments and are described
+below.
+It consists of *option objects*, *argument objects*, and *command  objects*.
+The depth of this object,
+since command objects in this array may themselves define command objects in
+their `arguments`,
+may not exceed 15.
```
poettering, line 153: i find it a bit weird that an array called `arguments` contains objects of which only some are considered `argument objects`.

Maybe rename `arguments` array to `parameters`? or alternatively rename `argument objects to `positional objects`?

(also, i know this is bikeshedding, but i wonder if `option` wouldn't be better called `switch`)

specs/cli_introspection.md
behrmann: Indeed, it was missing. Added. Though looking at it I noticed that further down I said it could be omitted, so I fixed that inconsistency, making it non-required with a default of `optional`.

specs/cli_introspection.md
behrmann: That's fine by me. The choice comes from directly taking this from getopt. 

I'm not entirely sold on `"no"`, though. It does follow from `no_argument` directly, so that makes sense, but "no" awfully close to a bool and if anybody made the mistake of using YAML, this could conceivable become one. `"none"` has a similar problem.

If anybody has an idea for a better name, I'm open to that, but for now I dropped `_argument` everywhere.

specs/cli_introspection.md
behrmann: name being an internal handle is exactly the idea, yes. Added a sentence to that regard and added the missing `valueName` documentation.

specs/cli_introspection.md
behrmann: > Maybe rename arguments array to parameters? or alternatively rename argument objects to positional objects`?

`parameters` seems a fine change to me, to make it one fewer `arguments` all over the place, but I think it muddies the water a bit, because in that context we'd still be talking about positional arguments and optional arguments. 

How about `optionalArgument` and `positionalArgument` for the objects?

> (also, i know this is bikeshedding, but i wonder if option wouldn't be better called switch)

I think switch as a term is too niche. I don't find switch in the bash or getopt man page, Python's argparse documentation calls everything an argument (betraying one of my influences :) ), the go parser I looked at calls these things options, so does Rust's clap. The latter *does* call binary options (`--foo`, `--no-foo`) switches and seemingly the Ruby option parser I looked at (with a *very* cursory look) uses switch more widely, but since everything, following `getopt` calls this whole field option parsing, I'd rather stick with something close to that.

behrmann:
> * Main help shows the commands and their summary
> * Command-specific help is `Usage: apt <command> ...` (possibly multiple lines) followed by the long description.
>
> I can map this to the spec to map usage to `arguments`, `help` for the short description, and `abstract` the long one.

Yes, so that's generally the idea that the usage, like `apt [opt] <command> …` comes from the name of the command and then options the `valueName` of arguments. For the discussion of `help` and `abstract` see the most recent commit. 
 
> One of the things that we likely can't address well here is multi-mode commands, notably we have:
> 
> ```
> Usage: apt [options] why <package>
>        apt [options] why-not <package>
> 
> Explain why a package is or is not installed. The why command
> shows why an installed package is present; why-not shows why a
> package cannot be installed. Takes a single package name.
> ```
> 
> And similarly say `install` and `remove` - it often makes sense to document them together if they have the same flags but inverse behavior. We'd instead need to repeat ourselves.
> 

My question here is: While it makes sense to group them for human-intended output, I don't think this necessarily needs to be grouped for tool-oriented output, e.g. I think it would be fine to describe both `why` and `why-not` as separate commands and include the same help text in both `Command` objects.

That being said, since a `Command` has an array of names, it could conceivably make sense to have `why` and `why-not` be names of the same command..

> One thing that could be more strictly defined to be useful is the `help` field - I'm thinking about `fish` completions which show a short description next to each item. The field is a bit loosely defined and reads like I could add a second-line, which sometimes is useful for `--help`, say:
> 
> ```
> --foo   Do bar. 
>             This is not the best thing in the world, consider using --bar instead.  # would not want this in fish
> ```
> 

Mhm, that kinda makes sense to me, we could make `help` an array, too, though that would weaken the distinction between `help` and `abstract` somewhat. 

My point of view is, that there are the following scenarios for some form of help text
- usage information, e.g. when a command is called without help, with no or too few arguments; this usually is a single line
- longer help output, e.g. when a command is called with `--help`
- some summary information, e.g. when a command appears in a list, e.g. as a subcommand in above help output
- some summary information shown during completion in certain shells like fish and zsh

In the current version, the second maps onto `abstract` and `postscript`, and I throw the first, third and fourth into `help` bucket. I'd be open making this more fine-grained, though considering that people are difficult to get to write documentation in the first place, I wonder how often such fine-grained distinction would be useful.

> Finally, option groups are often shared for us by up to 10 verbs, having to repeat ourselves is a bit of an overhead but it's likely not the end of the world.

I think the repetition there is actually a feature, annoying as it is, since commands shouldn't need information from their parents objects.

behrmann:
I pushed a few more fixes adding `valueName` were it was missing (but already used in the example at the bottom), added the `id` field as described in [this comment](https://github.com/uapi-group/specifications/pull/223#discussion_r4046313072), changed `default` to `isDefault` and added `synopsis` in line with https://github.com/systemd/systemd/pull/43790. I've also added a section on how to do custom extensions if necessary, since not everything from that PR is so widely useful that it needs to be here.

specs/cli_introspection.md
behrmann: I think it makes sense for them to have their own, e.g. git just accepts everything as a verb that's `git-foo` and there are third-party projects hooking into that and providing git subcommands. In that sense I think it makes sense to have this, since a top-level command that uses such an extension mechanism could still generate a global CLI introspection by calling `--introspect-cli` (or similar) on all of its extensions.

specs/cli_introspection.md
```diff
@@ -0,0 +1,674 @@
...
+
+### Command objects
+
+The command object is recursively defined as follows.
+
+```json
+{
+  "type": "command",
+  "id": "name1",
+  "names" : ["name1", "name2", …],
```
keszybz, line 97: I'm not convinced that "id" is needed. If it's an internal implementation detail (as the description defines it), then it should not be listed here. In almost all cases, the first name is completely fine as the id, or the id can derived from it.

I don't see why we should define something that has no immediate use and also duplicates the existing information.

specs/cli_introspection.md
```diff
@@ -0,0 +1,674 @@
...
+e.g. for single line usage information.
+
+### Argument objects
+
+Argument objects describe positional arguments that are not verbs.
+
+```json
+{
+  "type": "argument"
+  "id": "filename",
```
keszybz, line 201: Here "id" is useful, because otherwise the object is unnamed…

specs/cli_introspection.md
```diff
@@ -0,0 +1,674 @@
...
+`isDeprecated` is a boolean describing whether this option has been deprecated.
+
+### Option Objects
+
+Option objects describe optional arguments.
+
+```json
+{
+  "type": "option"
+  "id": "help",
```
keszybz, line 244: Also here, I'd make "id" optional. It's completely fine to derive the identifier from the names array. (E.g. take first long option or whatever. It's up to the consumer to decide if they need to.)

specs/cli_introspection.md
behrmann: I agree that using the first name as id is usually perfectly fine and should be used for commands and options, but in general I'd assume this JSON not to be handwritten, but generated, so I think it's okay to keep it mandatory.

Where I think `id` would be particularly useful would be in grouping options and arguments (e.g. for conflicting options or options that require other options as we have in ukify). I think it could work quite naturally to namespace the id field with dots or colons or double colons.

specs/cli_introspection.md
septatrix: How about epilog and prolog? The `script` naming seems quite confusing in this context

keszybz: Maybe "header" and "footer"? Sounds more down-to-earth.

septatrix:
If this is really supposed to become a universal standard it seems like it must get a decent bit more flexible even for quite common scenarios.

1. Currently there seems to be mention on how to support nested commands (like `git remote add` or `ip address show`).
2. There is no way to specify which flag applies to which command apart from general sections. This is also my biggest gripe with systemd's help output. Sure, I can look at `systemctl --help` but it does not tell me anything about which flags apply to `systemctl show`, or which commands repect `--output=json`.

And that is still excluding crazy shenanigans like GStreamer's `gst-launch` or `ffmpeg`

septatrix:
Oh and on another note: Many argument-parsing libraries which provide tab completion support, work by allowing to invoke the command with the currently spelled out flags/options and passing in another special flag (or environment variable) and then returning the list of feasible options. That is still quite limiting compared to e.g. fish's tab completion which supports showing a help text next to those completions and e.g. short+long form for flags. I think that this usecase could also be well supported by this spec without too many changes as it could mostly just reuse the schema of the `arguments` array I think

keszybz:
> If this is really supposed to become a universal standard it seems like it must get a decent bit more flexible even for quite common scenarios.

I don't disagree, but:

>     1. Currently there seems to be mention on how to support nested commands (like `git remote add` or `ip address show`).

`command` is recursive. This was added to support nested commands.

>     2. There is no way to specify which flag applies to which command apart from general sections.

The options that are listed at a certain level are assumed to apply to that command. Nested commands have their own option and subcommand sets. There is no mechanism to specify that a certain option can only be used with some subset of verbs… Maybe this could be added, but I'd leave that as an extension for later.

> This is also my biggest gripe with systemd's help output. Sure, I can look at `systemctl --help` but it does not tell me anything about which flags apply to `systemctl show`, or which commands repect `--output=json`.

Yes, but that also changes all the time, e.g. we keep adding support for `--output=json` in more places. The ambiguity is partially intentional.

behrmann:
> If this is really supposed to become a universal standard it seems like it must get a decent bit more flexible even for quite common scenarios.
> 

That is true, I still have a few outstanding things that need typing up:

- a type for option separators like `--`, which need an ability to dispatch to option parsing for other commands, which would be useful for wrapper commands that run another one. Completion for this is something that fish supports.
- a way to refer back to existing options, e.g. systemctl will happily accept options after the verb and its arguments, which is currently not covered here, so a completion generator would not know what to do
- number and separator of arguments for options 

>     1. Currently there seems to be mention on how to support nested commands (like `git remote add` or `ip address show`).
> 

Happy to clarify this if this is not obvious as it might be lost among the discussion here (and also in the PR by @keszybz  that is open on this PR in my fork of the repo), but the schema is recursive, so a command can have commands as arguments, this is the intended mechanism to capture subcommands like git does them.

I think I'll try to to (a subset) of ip as another example, as that is a sufficiently complex one to stress test this.

>     2. There is no way to specify which flag applies to which command apart from general sections. This is also my biggest gripe with systemd's help output. Sure, I can look at `systemctl --help` but it does not tell me anything about which flags apply to `systemctl show`, or which commands repect `--output=json`.
> 

Yeah, I'm in that boat as well. :) This is mainly a difference of taste, for systemd tooling all options live in a single namespace 

> And that is still excluding crazy shenanigans like GStreamer's `gst-launch` or `ffmpeg`

For `gst-launch` I assume you mean the pipeline descriptions? Yeah, no idea there, will have to look at this, but I think it's also fair to acknowledge that not every DSL may be captured.

> Oh and on another note: Many argument-parsing libraries which provide tab completion support, work by allowing to invoke the command with the currently spelled out flags/options and passing in another special flag (or environment variable) and then returning the list of feasible options. 

That is intended to be captured by the `dynamic` attribute for options, but maybe something similar for arguments would make sense as well.

poettering:
> 1. Currently there seems to be mention on how to support nested commands (like `git remote add` or `ip address show`).

so my understanding is that the ability to have commands objects as one type of arguments in command objects is what open this up. i.e. the command object "ip" has a command object called "address" as parameter, which has a command object "show" as parameter, and so on.

but honestly, we should play around with this, and see if we can actually model a semi-complex real-life project's cmdline that follows this style with this model. just because i think it works it doesn't mean it actually does ;-)

poettering:
> * a way to refer back to existing options, e.g. systemctl will happily accept options after the verb and its arguments, which is currently not covered here, so a completion generator would not know what to do

this is not specific to systemd btw. it's a property of glibc getopt/getopt_long, which we just inherited (and which i think makes sense). stupidly, getopt/getop_long on bsd and musl don't do that. but a while back we stopped using libc-provided getopt_long and since then actually expose identical behaviour on glibc and musl.

poettering:
> For `gst-launch` I assume you mean the pipeline descriptions? Yeah, no idea there, will have to look at this, but I think it's also fair to acknowledge that not every DSL may be captured.

so i guess we could fundamentally model the gst-launch `!` thing via a command object with a name of `!`. Except that we currently don't allow references to commands we already have defined, which we'd need here. maybe we should double down on the `id` thin, and then allow this to be used to reference already defined objects multiple times.

ototh maybe this would become too complex, and we should maybe not model this from day 1?

poettering:
oh, one major think we should add to the spec i think: this describes the json document, but doesn't actually say how to get it.

i.e. systemd's tools currently expose "--introspect-cli" for this, in all our tools. I think we should make that official, i.e. and suggest that any tool should dump its introspection that way if it supports introspection.

and while we are at it, maybe we should also recommend file suffix to use? in #213 i used `.Uapi16Manifest` as recommended suffix. We could take inspiration from that, hence `.Uapi18CliIntrospection` or so?

(i sense we might eventually see generators that use a json file as input and generate a parser from it, hence we should recommend a way to name these files from day one?

septatrix:
> Happy to clarify this if this is not obvious as it might be lost among the discussion here (and also in the PR by @keszybz that is open on this PR in my fork of the repo), but the schema is recursive, so a command can have commands as arguments, this is the intended mechanism to capture subcommands like git does them.
> 
> I think I'll try to to (a subset) of ip as another example, as that is a sufficiently complex one to stress test this.

Yes I think that would be great. `git` might also be a contender for these nested things as an example of a well known CLI tool.

> > Oh and on another note: Many argument-parsing libraries which provide tab completion support, work by allowing to invoke the command with the currently spelled out flags/options and passing in another special flag (or environment variable) and then returning the list of feasible options.
> 
> That is intended to be captured by the `dynamic` attribute for options, but maybe something similar for arguments would make sense as well.

Currently the `dynamic` attribute mentions that outputs need to be a single value per line. My though was that instead of printing just the values (or as an alternative) it could instead generate rich output (as JSON-SEQ, an array, etc) which again may contain `help`, `value`, `category` etc. Take for example `git rev-parse` which in fish generates a tabular autocompletion where next to the value I can see "Local/Remote branch", "Head", "Remote alias", or "Tag" for those respective things and commit hashes with their first commit message line.

image

> and while we are at it, maybe we should also recommend file suffix to use? in #213 i used `.Uapi16Manifest` as recommended suffix. We could take inspiration from that, hence `.Uapi18CliIntrospection` or so?

I personally find those uppercase extensions quite ugly because they are used effectively nowhere else. I'd rather see `.uapi16-manifest`/`.uapi18-cli-introspection` or something along those lines

> (i sense we might eventually see generators that use a json file as input and generate a parser from it, hence we should recommend a way to name these files from day one?

Hm, most people I know and work with have a lot of reservations against code generators, I think the other way around will be more likely. However, I work mostly with higher level languages like Python, Rust and JS where introspection is a lot simpler/more powerful and things are done the other way 'round (generate specs from the code).

---

Another thing that might be worth taking a look into is to specify this as a JSON schema

PR #226, opened 2026-09-03
uki: add .mokkeys section

andrewdunndev:
Companion to systemd/systemd#43638, which adds the section to ukify and systemd-measure.


bluca:
What's the use case for this exactly? If one already has access to a key that can sign and verify a UKI, then by definition the same key can be used for everything else too

Foxboron:
@bluca See the proposal on the systemd repo.

https://github.com/systemd/systemd/issues/43539

bluca:
Yeah that makes no sense either and doesn't really explain anything. Why can't the key used to sign the image be used to sign the modules?

andrewdunndev:
Closing this. The use case was an out-of-tree module signed with my own key on a distribution kernel I didn't build, on machines where I own `db` and run no shim; the reasons I'm dropping the `.mokkeys` approach, and what I found about the MOK table along the way, are in one comment on the systemd PR: https://github.com/systemd/systemd/pull/43638#issuecomment-5611922484. Thanks for the questions; they were the right ones. Thanks also to @Foxboron for the pointer.

PR #227, opened 2026-09-04
vmclock: Update to v1.1 with fixes for offsets, flags and virtio-rtc references

dwmw2:
The offset column for the structure was inconsistent. Arguably it should have been generated automatically with offsetof(), but wasn't. The time_maxerror_nanosec field is 64 bits, which puts vm_generation_count at 0x68 not 0x64. The size is therefore 0x70.

More importantly, the feature flags table omitted the TIME_MONOTONIC flag, which has always been at bit 7 in the Linux implementation since it was first merged there in 2024. Then the newer VM_GEN_COUNTER_PRESENT and NOTIFICATION_PRESENT bits got added in the wrong place in the spec too (at 7,8 while Linux had them at 8,9). Retcon those to match Linux, whose header file is also used in QEMU and *is* the canonical reference to follow at this point.

Rename vm_generation_count to vm_generation_counter throughout, matching the header and the spec's own flag description.

The tai_offset_sec field is correctly described as the offset from TAI to UTC, and the Linux driver consistently implements that (tai_adjust() subtracts it from UTC to yield TAI). But the example value given was 37, which is TAI - UTC. The value as currently populated is -37; spell out the convention since the commonly-quoted offset has the other sign.

Align the time_type list with the Linux header: the value is VMCLOCK_TIME_MONOTONIC not VMCLOCK_MONOTONIC, and values 3 and 4 are reserved for the (unsupported) smeared clock types from virtio-rtc.

Replace the placeholder virtio-rtc link with the published standard (Virtio 1.4, section 5.23 "RTC Device"), and update the embedded Device Tree binding to match the version merged into Linux, which lives in the ptp/ schema directory and refers back to this document.

The leap_second_smearing_hint field was described as corresponding to the draft virtio-rtc "subtype" field, which did not survive into the published standard; there, smeared clocks became distinct clock types and the leap_second_smearing field describes the exposed clock itself, which for VMClock is never smeared. Describe the relationship to the published standard instead, noting that the NOON_LINEAR and UTC_SLS values still match but value 0 does not.

Reported-by: Torben Hansen <htorben@amazon.com>

### DIFF
--- a/specs/vmclock.md
+++ b/specs/vmclock.md
@@ -2,7 +2,7 @@
 title: UAPI.13 VMClock
 category: Concepts
 layout: default
-version: 1.0
+version: 1.1
 SPDX-License-Identifier: CC-BY-4.0
 weight: 13
 aliases:
@@ -15,6 +15,7 @@ aliases:
 | Version | Changes         |
 |---------|-----------------|
 | 1.0     | Initial Release |
+| 1.1     | Fix vm_generation_counter offset and flag bit numbers; add missing VMCLOCK_FLAG_TIME_MONOTONIC; fix sign of tai_offset_sec example value; update references to published virtio-rtc standard |
 
 The requirements for accurate synchronisation of application clocks against
 real wallclock time are becoming ever more demanding. Increasingly cloud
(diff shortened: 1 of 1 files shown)


keszybz (approved):
LGTM.

keszybz:
I'm not an expert on any of this, but the individual changes and references are all correct afaict. Let's wait for @brauner's review though too.

brauner (approved):

PR #228, opened 2026-09-07
Add UAPI.20: codify varlink IPC in an uapi spec

poettering:
I think we should formalize Varlink a bit better, in a proper spec, as part of UAPI.

here's an initial version, derived from the text on varlink.org, and updated in various way.

I left out the varlink registry for now (we can add this later), but did include some notes on applying Varlink on AF_UNIX and on HTTP

### DIFF
--- a/specs/varlink.md
+++ b/specs/varlink.md
@@ -1,0 +1,1399 @@
+---
+title: UAPI.20 Varlink IPC
+category: Concepts
+layout: default
+version: 0.1
+SPDX-License-Identifier: CC-BY-4.0
+weight: 20
+aliases:
+- /UAPI.20
+- /20
+---
+
+# UAPI.20 Varlink IPC
+
+| Version | Changes         |
+|---------|-----------------|
+| 0.1     | Initial release |
+
+Varlink is an interface description format and IPC protocol that aims
+to make services accessible to both humans and machines in the
+simplest feasible way.  A Varlink interface combines what the classic
+UNIX command line options, `STDIN`/`STDOUT`/`STDERR` text formats, man
+pages and service metadata provide, and makes the equivalent available
+over a single file descriptor.
+
+Varlink is plain-text, type-safe, discoverable, self-documenting,
+remotable, testable and easy to debug. It is accessible from any
+programming environment: clients do not require complex modules or
+libraries, existing JSON and local socket communication facilities are
+sufficient for basic operation.
+
+This specification consolidates the protocol documentation originally
+published at [varlink.org](https://varlink.org/), and extends it with
+a definition of how the protocol is bound to specific transports:
+`AF_UNIX` stream sockets (including how such sockets are marked so
+that they can be recognized as Varlink sockets) and HTTP (in a simple
+request/response mode and in a WebSocket mode).
+
+## Design Principles
+
+* **Simplicity.** Varlink aims to be as simple as possible. It is not
+  specifically optimized for anything else but ease-of-use and
+  maintainability.
+
+* **Discoverability.** Varlink services describe themselves, with a
+  machine-readable interface definition and human-readable
+  documentation. The documentation is provided to clients as
+  structured data on a well-known service interface.
+
+* **Remotability.** It should be easy to forward, bridge and redirect
+  Varlink interfaces over any connection-oriented transport. Varlink
+  should minimize exposure to side-effects of local APIs. Interactions
+  are designed to be simple messages on a network, not carrying direct
+  references to locally stored file paths or needlessly relying on
+  passing file descriptors.
+
+* **Focus.** Varlink is the protocol and the definition of interfaces,
+  but it does not define or provide any significant functionality
+  itself.
+
+* **Errors.** Varlink errors carry enough information to be consumed
+  by machines and automated rules.
+
+Varlink uses direct connections, has no central message handling component,
+and is hence easily debugged, secured, isolated and tested. Everything is
+readable: Varlink uses plain text messages, has no magic numbers and no
+unnamed values, and is easily debuggable with tools such as `strace`, `netcat`
+or `jq`.
+
+## Terminology
+
+* An **interface** is a named collection of types, methods and errors,
+  described in the Varlink interface definition language (see below).
+
+* A **service** is a program that implements one or more interfaces and
+  accepts connections from clients.
+
+* A **client** is a program that connects to a service and invokes methods.
+
+* A **connection** is a bidirectional, ordered, reliable byte stream between a
+  client and a service, established over a *transport*.
+
+* A **message** is a single JSON object exchanged over a connection, either a
+  *call* (client to service) or a *reply* (service to client).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in BCP
+14 [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) [RFC
+8174](https://www.rfc-editor.org/rfc/rfc8174) when, and only when,
+they appear in all capitals, as shown here.
+
+## Interface Definition
+
+A Varlink interface has a reverse-domain name and specifies which methods the
+interface implements. Each method has named and typed input and output
+parameters. Complex types can be aliased with the `type` keyword to allow
+reusing them and to make method signatures easier to read. The interface also
+specifies the errors that may be returned from its method calls.
+
+Everything can be documented by adding a comment immediately before it.
+
+Interface definition files are named after the interface they implement with
+the suffix `.varlink`, e.g. `org.example.ftl.varlink`.
+
+### Example
+
+```
+# Interface to jump a spacecraft to another point in space.
+# The FTL Drive is the propulsion system to achieve
+# faster-than-light travel through space. A ship making a
+# properly calculated jump can arrive safely in planetary
+# orbit, or alongside other ships or spaceborne objects.
+interface org.example.ftl
+
+# The current state of the FTL drive and the amount of
+# fuel available to jump.
+type DriveCondition (
+  state: (idle, spooling, busy),
+  tylium_level: int
+)
+
+# Speed, trajectory and jump duration is calculated prior
+# to activating the FTL drive.
+type DriveConfiguration (
+  speed: int,
+  trajectory: int,
+  duration: int
+)
+
+# The galactic coordinates use the Sun as the origin.
+# Galactic longitude is measured with primary direction
+# from the Sun to the center of the galaxy in the galactic
+# plane, while the galactic latitude measures the angle
+# of the object above the galactic plane.
+type Coordinate (
+  longitude: float,
+  latitude: float,
+  distance: int
+)
+
+# Monitor the drive. The method will reply with an update
+# whenever the drive's state changes
+method Monitor() -> (condition: DriveCondition)
+
+# Calculate the drive's jump parameters from the current
+# position to the target position in the galaxy
+method CalculateConfiguration(
+  current: Coordinate,
+  target: Coordinate
+) -> (configuration: DriveConfiguration)
+
+# Jump to the calculated point in space
+method Jump(configuration: DriveConfiguration) -> ()
+
+# There is not enough tylium to jump with the given
+# parameters
+error NotEnoughEnergy ()
+
+# The supplied parameters are outside the supported range
+error ParameterOutOfRange (field: string)
+```
+
+### Keywords
+
+| Keyword     | Description                                                                                                 |
+|-------------|-------------------------------------------------------------------------------------------------------------|
+| `interface` | The name of the interface in reverse-domain name notation. It MUST be the first keyword in the description. |
+| `type`      | A custom type: the definition of an object/record/structure, or of an enumeration.                          |
+| `method`    | A method with input and output parameter objects.                                                           |
+| `error`     | The name of an error, and an optional data object describing the error.                                     |
+
+Types, methods and error names MUST be unique identifiers within an interface
+definition; they share the same namespace. They are referenced by their
+fully-qualified reverse-domain name. The method `GetInfo()` in the interface
+`org.example.test` is called `org.example.test.GetInfo`.
+
+### Names
+
+* **Interface names** use reverse-domain notation. They consist of at least
+  two dot-separated components. Each component starts with a letter (or, for
+  all but the first component, a digit), and continues with letters, digits
+  and non-leading, non-trailing hyphens. See the grammar below for the exact
+  syntax.
+
+* **Type, method and error names** start with an uppercase ASCII letter and
+  continue with ASCII letters and digits.
+
+* **Field names** (of structures and enumerations, and of method parameters)
+  start with an ASCII letter and continue with ASCII letters and digits. An
+  underscore (`_`) may appear between two such characters, but not at the
+  beginning or the end of the name, and not doubled.
+
+### Types
+
+With `type` a structure of fields (an object) or an enumeration can be
+defined. Following the `type` keyword the type name has to be specified.
+Enclosed in `()` the declaration of the various fields follows. Field
+declarations specify a field name followed by a colon and a type. Fields are
+separated by commas.
+
+```
+type MyType (
+   example_bool: bool,
+   example_int: int,
+   example_float: float,
+   example_string: string,
+   example_object: object,
+   example_any: any,
+   example_enum: (one, two, three),
+   example_struct: (first: int, second: string),
+   example_array: []string,
+   example_dictionary: [string]string,
+   example_stringset: [string](),
+   example_nullable: ?string,
+   example_nullable_array_struct: ?[](first: int, second: string),
+   example_other_type: MyOtherType
+)
+```
+
+| Type           | Keyword    | Description                                                                                                          | Example                        |
+|----------------|------------|----------------------------------------------------------------------------------------------------------------------|--------------------------------|
+| Boolean        | `bool`     | `true` or `false`                                                                                                    | `flag: bool`                   |
+| Integer        | `int`      | Signed integer, implementation-specific size, usually 64-bit                                                         | `whole: int`                   |
+| Floating Point | `float`    | Implementation-specific size, usually IEEE 754 double                                                                | `half: float`                  |
+| String         | `string`   | Unicode text                                                                                                         | `name: string`                 |
+| Object         | `()`       | Structure of named, typed fields, separated by comma                                                                 | `(first: int, second: string)` |
+| Enum           | `()`       | List of names without types, separated by comma                                                                      | `(this, that, other)`          |
+| Map/Dictionary | `[string]` | Map of values of one type, keyed by unique strings                                                                   | `[string]int`                  |
+| Array          | `[]`       | Ordered list of values of one type                                                                                   | `[]int`                        |
+| Nullable       | `?`        | Nullable type ("maybe")                                                                                              | `?string`                      |
+| Foreign Object | `object`   | Foreign, untyped/raw JSON object                                                                                     | `payload: object`              |
+| Any            | `any`      | Any value: Boolean, Integer, Floating Point, String, (Foreign) Object, Enum, Map/Dictionary, Array, but not Nullable | `value: any`                   |
+
+A *set of strings* can be expressed as a map with an empty structure as value
+type, i.e. `[string]()`.
+
+Structures and enumerations may be declared anonymously inline (as in the
+example above), or given a name via `type` and referenced by that name.
+
+### Methods
+
+Method names start with an uppercase character and continue with alphanumeric
+characters. A method specifies an input and an output parameter object:
+
+```
+method Add(a: int, b: int) -> (sum: int)
+method Foo(a: int, b: MyType) -> (bar: MyOtherType, baz: float, more: (i: int, f: float, s: string))
+```
+
+Both the input and the output parameter list may be empty, expressed as `()`.
+
+### Errors
+
+Error names start with an uppercase character and continue with
+alphanumeric characters, and optionally carry parameters describing
+the error. The error is identified by its fully-qualified
+name. Varlink calls SHOULD only return well-defined errors specified
+in the interface description (or in the `org.varlink.service`
+interface, see below).
+
+```
+error UnknownAction(action: string, more_data: DetailedError)
+```
+
+### Documentation Comments
+
+All whitespace (space, `\t`, `\r`, or `\n`) is ignored, except inside
+comments, which are started with `#` and extend to the end of the line, and
+except that the interface name and each member declaration MUST be followed
+by a line break (see the grammar below).
+
+Comments that start on a new line and are immediately followed by
+`interface`, `method`, `type` or `error` are documentation for that following
+declaration. These comments may span multiple lines. Comments immediately
+preceding a field declaration inside a structure, enumeration or parameter
+list are documentation for that field.
+
+A single space following the `#` character is conventionally used as
+separator and is not part of the documentation text.
+
+Because Varlink interfaces are retrieved as plain text at runtime (see
+`org.varlink.service.GetInterfaceDescription` below), the documentation
+comments are transmitted to clients together with the interface definition,
+and thus available for introspection and for generating user-visible help.
+
+### Grammar
+
+```
+interface
+        "interface" interface_name members
+
+members
+        member
+        member members
+
+member
+        type_alias
+        method
+        error
+
+type_alias
+        "type" name struct
+        "type" name enum
+
+method
+        "method" name struct "->" struct
+
+error
+        "error" name struct
+
+type
+        element_type
+        "[]" type
+        "[string]" type
+        "?" element_type
+        "?" "[]" type
+        "?" "[string]" type
+
+element_type
+        struct
+        enum
+        name
+        "bool"
+        "int"
+        "float"
+        "string"
+        "object"
+        "any"
+
+struct
+        "(" struct_fields ")"
+        "(" ")"
+
+struct_fields
+        struct_field
+        struct_field "," struct_fields
+
+struct_field
+        field_name ":" type
+
+enum
+        "(" enum_fields ")"
+
+enum_fields
+        field_name
+        field_name "," enum_fields
+
+interface_name
+        [A-Za-z]([-]*[A-Za-z0-9])*(\.[A-Za-z0-9]([-]*[A-Za-z0-9])*)+
+
+name
+        [A-Z][A-Za-z0-9]*
+
+field_name
+        [A-Za-z](_?[A-Za-z0-9])*
+```
+
+### Parsing Expression Grammar
+
+The following is the same grammar expressed as a Parsing Expression Grammar
+(PEG), including whitespace and comment handling:
+
+```
+whitespace /* Modeled after ECMA-262, 5th ed., 7.2. \v\f removed */
+    = [ \t\u{00A0}\u{FEFF}\u{1680}\u{180E}\u{2000}-\u{200A}\u{202F}\u{205F}\u{3000}]
+
+eol_r /* Modeled after ECMA-262, 5th ed., 7.3. */
+  = "\n"
+  / "\r\n"
+  / "\r"
+  / "\u{2028}"
+  / "\u{2029}"
+
+comment
+    = "#" [^\n\r\u{2028}\u{2029}]* eol_r
+
+eol
+    = whitespace* eol_r
+    / comment
+
+_
+    = whitespace / comment / eol_r
+
+field_name
+    = [A-Za-z]('_'?[A-Za-z0-9])*
+
+name
+    = [A-Z][A-Za-z0-9]*
+
+interface_name
+    = [A-Za-z]([-]* [A-Za-z0-9])* ( '.' [A-Za-z0-9]([-]*[A-Za-z0-9])* )+
+
+dict
+    = "[string]"
+
+array
+    = "[]"
+
+maybe
+    = "?"
+
+element_type
+    = "bool"
+    / "int"
+    / "float"
+    / "string"
+    / "object"
+    / "any"
+    / name
+    / venum
+    / vstruct
+
+type
+    = element_type
+    / maybe element_type
+    / array type
+    / dict type
+    / maybe array type
+    / maybe dict type
+
+venum
+    = '(' ( _* field_name ** ',' ) _* ')'
+
+argument
+    = _* field_name _* ':' _* type
+
+vstruct
+    = '(' ( argument ** ',' ) _* ')'
+
+vtypedef
+    = "type" _+ name _* vstruct
+    / "type" _+ name _* venum
+
+error
+    = "error" _+ name _* vstruct
+
+method
+    = "method" _+ name _* vstruct _* "->" _* vstruct
+
+member
+    = _* m:method
+    / _* t:vtypedef
+    / _* e:error
+
+interface
+    = _* "interface" _+ interface_name eol ( member ++ eol ) _*
+```
+
+### Interface Evolution
+
+Varlink interfaces do not have a version number, they only have a feature set
+described in detail by the interface definition, which is part of the wire
+protocol. Clients can fully introspect a service and figure out which features
+the interface supports; this is more expressive than any simple numbering
+scheme.
+
+Varlink does not use positional parameters or fixed-size objects in its
+interface definition or on the wire: all parameters are identified by their
+name and can be extended later. Interfaces SHOULD be extended in a backwards
+compatible way:
+
+* New fields SHOULD be added as nullable (`?`) fields, both in structures and
+  in method input/output parameters. The result of a method, when passed
+  parameters without the new optional fields, SHOULD be the same as in older
+  versions. The expected behavior for omitted fields SHOULD be documented.
+
+* New methods, types and errors may be added. New enumeration values may be
+  added to enumerations used in method input parameters; adding values to
+  enumerations that appear in replies is only backwards compatible if clients
+  are prepared to handle unknown values, which SHOULD be documented.
+
+* Removing or incompatibly changing fields, types, errors, enumeration values
+  or methods is not backwards compatible and SHOULD be avoided. If an
+  interface must be changed incompatibly, it SHOULD be released under a new
+  name, commonly by adding a number to the interface name, e.g.
+  `org.example.interface2`.
+
+Services SHOULD refuse unrecognized fields in method calls with an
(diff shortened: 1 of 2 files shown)


specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+A Varlink interface combines what the classic UNIX command line options,
+`STDIN`/`STDOUT`/`STDERR` text formats, man pages and service metadata provide,
+and makes the equivalent available over a single file descriptor.
+
+Varlink is plain-text, type-safe, discoverable, self-documenting, remotable,
+testable and easy to debug. It is accessible from any programming environment:
+clients do not require complex modules or libraries, existing JSON and socket
+communication facilities are sufficient for basic operation.
+
+This specification consolidates the protocol documentation originally
```
bluca, line 32: if instead of "consolidates" you say "embraces" here we could do the funniest thing

bluca:
Aren't you an owner in the varlink org? Why here instead of there?

poettering:
> Aren't you an owner in the varlink org? Why here instead of there?

Well, because UAPI is kind of an established standards body, I guess, and varlink.org is not really. I think it would make sense to maybe host a copy of the spec on varlink.org, but I think uapi is a better home for a spec.

I mean, would expect to look for the spec for HTTP on http.org? I'd check the IETF first, for its RFC standardizing scheme

chergert:
One thing I'd like to consider is when you're in a situation that you only have a single socket you can use, and what to multiplex over it (say USB device connection). In `varlink-glib` I ended up creating a simple muxer. Obviously most situations can just create a socket, so this is more about when that is not an option.

https://libgvarlink-1f981c.pages.gitlab.gnome.org/varlink-glib-1/protocol-muxer.html

bluca:
> > Aren't you an owner in the varlink org? Why here instead of there?
> 
> Well, because UAPI is kind of an established standards body, I guess, and varlink.org is not really. I think it would make sense to maybe host a copy of the spec on varlink.org, but I think uapi is a better home for a spec.
> 
> I mean, would expect to look for the spec for HTTP on http.org? I'd check the IETF first, for its RFC standardizing scheme

Ok, makes sense

poettering:
> One thing I'd like to consider is when you're in a situation that you only have a single socket you can use, and what to multiplex over it (say USB device connection). In varlink-glib I ended up creating a simple muxer. Obviously most situations can just create a socket, so this is more about when that is not an option.

> https://libgvarlink-1f981c.pages.gitlab.gnome.org/varlink-glib-1/protocol-muxer.html

yeah, i guess one day we should look into having a scheme you could apply to rs232 or usbstream that adds the thinnest concept of multiplexing we can get away with, but maybe let's do that in a later step, and first start out by formalizing what we already have.

I spend some time on looking for established multiplexer protocols for this kind of use, but found nothing I particularly liked.

(two things looking at your multiplexer spec that pop into my head though when reading through it: seems strange to add a binary size field for the multiplexer frame size, given that raw varlink tries to avoid that and goes for NUL frame separators instead, might be worth sticking to doing things that way. dunno. which brings me to the other thing: maybe the solution is to extend varlink's message header instead. i.e. add an extra field to the regular message header named `stream` which carries a stream identifier. return stream connection errors as regular varlink errors. treat the regular stream as stream with id "0".)

zeenix:
> goes for NUL frame separators instead

Yet another reason this was a bad idea IMO. If a binary header must be avoided, there could also an outer JSON wrapper like this:

```json
{
   size: 24, // Binary size of the `msg` value.
   msg: {
      // The actual message.
   }
}
```

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+  speed: int,
+  trajectory: int,
+  duration: int
+)
+
+# The galactic coordinates use the Sun as the origin.
+# Galactic longitude is measured with primary direction
+# from the Sun to the center of the galaxy in the galactic
+# plane, while the galactic latitude measures the angle
+# of the object above the galactic plane.
```
zeenix, line 135: * Nitpick: Can we use the same line width everywhere in this document. I suggest 100 chars.
* Good to have this documented so nerds don't complain about the 2D nature of this. 😆

chergert:
Injecting the muxer channel into the `{}` is sort of like VLAN tagging where you add 4 bytes after the source mac. But in this case, it requires the muxer to also potentially parse/strip fields from the caller. That would push me more towards an outer wrapper if we were not doing framing.

But we can always discuss this on a specific topic in the future. I just wanted to put it out there. And yes, rs232/usbstream are my primary use cases.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+### Nullable Fields and Absent Fields
+
+For a nullable field (`?T`), the following two encodings MUST be treated as
+equivalent by receivers: the field being absent from the enclosing JSON
+object, and the field being present with the value `null`. Note that for
+fields of type `?[]T` or `?[string]T` an empty array or object respectively is
+*not* equivalent to `null`: an empty collection is a valid non-null value.
+
+Senders SHOULD omit fields whose value is `null` rather than encoding them
+explicitly.
```
zeenix: Thanks for adding this but I think this should be a MUST. See https://github.com/varlink/varlink.github.io/issues/36 for discussion and reasoning. @poettering and please don't say "fix your language". That's not super helpful and what I'm asking for is IMO good thing in general as it will allow (especially auto-generated from IDL) parsers to be simpler.

zeenix (changes requested):
I skimmed through mostly and read some specific parts in detail. It all looks good to me. Thanks for doing this as we really need this.

mvo5:
> > One thing I'd like to consider is when you're in a situation that you only have a single socket you can use, and what to multiplex over it (say USB device connection). In varlink-glib I ended up creating a simple muxer. Obviously most situations can just create a socket, so this is more about when that is not an option.
> 
> > https://libgvarlink-1f981c.pages.gitlab.gnome.org/varlink-glib-1/protocol-muxer.html
>
> yeah, i guess one day we should look into having a scheme you could apply to rs232 or usbstream that adds the thinnest concept of multiplexing we can get away with, but maybe let's do that in a later step, and first start out by formalizing what we already have.
> 
> I spend some time on looking for established multiplexer protocols for this kind of use, but found nothing I particularly liked.
[..]

Fwiw, in the varlink-http-bridge I'm experimenting with using a single connection to carry multiple varlink connections. There my thinking is to use the native http2 (h2) stream multiplexing. Probably a bit heavy for the other use cases but I wanted to mention it since h2 has all we need plus per-stream flow-control. 

One issue with idea of a json header for the stream is protocol upgrades: those would not work anymore after the upgrade in a multiplexed setup. So I think we need some "outer" frame.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+  are designed to be simple messages on a network, not carrying direct references to locally stored file
+  paths or needlessly relying on passing file descriptors.
+
+* **Focus.** Varlink is the protocol and the definition of interfaces, but it
+  does not define or provide any significant functionality itself.
+
+* **Errors.** Varlink errors carry enough information to be consumed by
+  machines and automated rules.
+
+* **Testing.** Making it easy to run automated tests against Varlink
```
mvo5: This feels a bit like fluff, I am not sure what it even means, but maybe it becomes clearer as I read on

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+
+Varlink connections can thus be turned into monitor connections, which are
+soft-negotiated between the client and the service. To avoid race conditions
+between the reading of the current state and subscribing to updates, a typical
+Varlink monitor method call first returns the current state and then sends
+changes in subsequent replies to the same request on the same connection.
+
+The IDL definition of a method that supports `more` MUST carry the combination of all possible replies as
+reply message definition, and MUST set `?` on all fields that may be absent in at least one valid reply of
+the series.
```
mvo5: Maybe add something here about the convention systemd uses to expose this in
the IDL, at least until https://github.com/varlink/varlink.github.io/issues/26
is resolved. systemd's IDL dump emits a marker line in the method's doc
comment, and tools like varlink-http-bridge parse it to know which methods
stream. Something like:

```
To let clients learn ahead of time whether a method supports or requires
`more`, the documentation comment of such a method SHOULD contain a line
consisting of exactly `[Supports 'more' flag]` or `[Requires 'more' flag]`,
e.g.


  # Enumerates all sessions.
  # [Requires 'more' flag]
  method ListSessions() -> (session: Session)


Implementations MAY use these lines to derive the method's `more` semantics
and SHOULD omit them from user-visible documentation.
```

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+likewise MUST NOT consume bytes beyond the terminator of the confirming reply.
+Clients SHOULD NOT send any data belonging to the upgraded protocol before
+having received the confirmation.
+
+This mechanism is useful for transferring larger amounts of data, or foreign
+data that is not reasonably encoded in JSON. It is conceptually similar to
+the WebSocket upgrade in HTTP.
+
+A server SHOULD reply with the `org.varlink.service.ExpectedUpgrade` error to a call of a method that
+requires the flag but did not set it.
```
mvo5: Same suggestion as above for "more" for procol upgrades. systemd already reads the magic doc comment here too.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+
+For socket-activated Varlink services managed by a service manager, the extended attributes should be set by
+the service manager when allocating them, and not be delayed until the service they are associated with
+ultimately get activated.
+
+## Transport: HTTP
+
+Varlink can be encapsulated in HTTP, in order to make services available remotely, in particular to
+environments where the `AF_UNIX` transport is not available, such as web browsers or hosts of virtual
+machines. This section specifies such an encapsulation. It is typically implemented by a proxy that accepts
```
mvo5: s/proxy/bridge/ here maybe? Given that the thing we have is the varlink-http-**bridge**

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+### URL Schema
+
+| Method | Path                                | Description                                                                           |
+|--------|-------------------------------------|---------------------------------------------------------------------------------------|
+| `POST` | `/call/{method}`                    | Invoke a method. The socket is derived from the method name, or given via `?socket=`. |
+| `POST` | `/call/{socket}/{method}`           | Invoke a method on an explicitly given socket.                                        |
+| `GET`  | `/sockets`                          | List the available sockets.                                                           |
+| `GET`  | `/sockets/{socket}`                 | Service information of one socket (as returned by `org.varlink.service.GetInfo`).     |
+| `GET`  | `/sockets/{socket}/{interface}`     | Details of one interface implemented by a socket, including its method names.         |
+| `GET`  | `/openapi/{socket}/{interface}`     | OpenAPI 3.1 description generated from the interface definition.                      |
```
mvo5: I'm not fully certain we should have this openapi endpoint as part of the spec just yet. Lets gather some experinece from users first. We have the code but little real-world experience with this.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+authenticate clients, unless the network is trusted by construction. Two
+mechanisms are defined:
+
+* **Mutual TLS.** The proxy is configured with a CA certificate, and requires
+  clients to present a certificate signed by it. Without a configured
+  certificate the proxy MAY generate a self-signed one, in which case clients
+  SHOULD pin the server's public key on first contact (in the manner of SSH
+  `known_hosts`), and refuse connections if it later changes.
+
+* **SSH key signatures.** Requests are signed with an SSH private key (Ed25519
```
mvo5: Same as the openapi comment above. I'm not sure we need to standardize this just yet. I think having a good mimimum for authentication (mTLS) is a must but beyond that I think we can (for now) leave this open to the implemtation of the varlink-http-bridge until we have more real-world experinece. I fully expect that we get api-key and jwt support soon here and it may well be that sshauth is a bit of a niche (or maybe not, its hard to say at this point).

mvo5 (commented):
Thanks for this ! Reads mostly great, some small inline comments for your consideration.

specs/varlink.md
mrc0mmand: I was just about to comment this and this is one of the things that makes writing an automatic Varlink fuzzer quite annoying.

There are already open RFEs on Valink's side to incorporate `oneway`, `upgrade`, and `more`  into the IDL:
  - https://github.com/varlink/varlink.github.io/issues/41
  - https://github.com/varlink/varlink.github.io/issues/42
  - https://github.com/varlink/varlink.github.io/issues/26

And having this would be great, since then we could get this information programmatically through the interface introspection, as we can already do with D-Bus annotations.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+error InvalidParameter (parameter: string)
+
+# Client is denied access
+error PermissionDenied ()
+
+# Method is expected to be called with 'more' set to true, but wasn't
+error ExpectedMore ()
+
+# Method is expected to be called with 'upgrade' set to true, but wasn't
+error ExpectedUpgrade ()
```
mrc0mmand, line 853: I'm not sure if the goal of this PR is to describe the status quo or the "ideal" implementation, but it would be great to extend this interface with a `ProtocolError` (or something like that), as currently there's no standardized way how to notify the other side that the current request caused some sort of protocol violation - currently, the other side just "hangs up", so distinguishing between a protocol error and a crash is a bit annoying, see https://github.com/systemd/systemd/issues/41282.

specs/varlink.md
poettering: still not grokking why allowing all three ways to encode an empty param list wouldn't be ok. All i am hearing is that serde can't parse that. but i don't get why this limitation of serde should leak into the spec, and why one cannot just fix serde to allow this.

would be great if someone could actually explain to me what the issue is supposed to be with allowing all three syntaxes for this.

the spec already says that generally absence of a field should be considered equivalent to be it set to null. and that absence/null should by understood as "doesn't apply" or "use defaults".

So if we apply this logic to the parameter list too and say that the empty parameter list is the default, then it's only systematic to allow null, absence and `{}` for it meaning the same thing.

specs/varlink.md
poettering: So i think we should start encoding this, but in a more formal way rather than in comments. but i'd really do that in a later revision of the spec.

putting this in a comment was me chickening out of actually extending the spec for this.

extending the spec probably means introducing some versioning of the IDL language, so that clients can ask for a certain version of the IDL with only features they support

specs/varlink.md
zeenix: Serde issue is just a symptom and it's not even a serde issue really (I can try to explain once again if you promise to listen this time). Other parsers can easily have the same issue.

I mentioned above that at least one advantage of mandating one specific way here: it can help simplify parser code. Did you read through the issue I linked to? There are good arguments there.

You can also ask yourself the opposite question: What's the harm in mandating this, other than you can see this as breaking backwards compatibility (which could be covered by a recommendation for implementation that care about that)?

It's best to make the spec strict and specific IMO.

specs/varlink.md
bachorp: Tangential question: Would you consider having also a structured format for introspection? This could simply be a JSON representation of a parsed IDL file.

With that, the client (and client developer) would not be burdened with parsing and could utilize introspection more smoothly. A structured format is also usually easier to extend backwards-compatibly (adding a language keyword vs. an optional object key).

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+method Add(a: int, b: int) -> (sum: int)
+method Foo(a: int, b: MyType) -> (bar: MyOtherType, baz: float, more: (i: int, f: float, s: string))
+```
+
+Both the input and the output parameter list may be empty, expressed as `()`.
+
+### Errors
+
+Error names start with an uppercase character and continue with alphanumeric
+characters, and optionally carry an error object describing the error. The
```
bachorp: "optionally carry an error object" can be misunderstood. The object is optional on the wire, but `()` is required in the interface description.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+
+* A **connection** is a bidirectional, ordered, reliable byte stream between a
+  client and a service, established over a *transport*.
+
+* A **message** is a single JSON object exchanged over a connection, either a
+  *call* (client to service) or a *reply* (service to client).
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
```
aphedges: RFC 2119 was updated by [RFC 8174](https://www.rfc-editor.org/info/rfc8174/) in 2017, which revises this text to clarify that key word capitalization is significant.

Given that this specification uses both `MAY` and `may`, I feel that clarification would probably be helpful. It would likely require adopting the new text and making sure the case of all key words is as intended.

aphedges (changes requested):

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+socket explicit; this is the form the generated OpenAPI descriptions use.
+
+The proxy issues the call over a Varlink connection to the socket (the
+connection MAY be reused for subsequent requests, subject to the ordering
+rules of the protocol) and maps the reply to the HTTP response as follows:
+
+* **Success.** Status `200 OK`, with the reply's output parameters as JSON
+  body — again the value of the `parameters` member rather than the reply
+  object, i.e. the reply is *unwrapped*. An absent `parameters` member is
+  returned as `{}`.
```
Skyb0rg007, line 1194: While HTTP doesn't support Varlink's `oneway` semantics entirely, we could still add some way to signal that the client isn't interested in the response, and have the server respond with `204 No Content` + empty body. That would be useful for supporting expensive Varlink methods, where a `oneway` flag lets the server to perform the actual operation asynchronously.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+$ curl -s -H "Accept: application/json-seq" -H "Content-Type: application/json" \
+    https://host:1031/call/io.systemd.UserDatabase.GetUserRecord \
+    -d '{"service":"io.systemd.Multiplexer"}' | jq --seq
+```
+
+#### Limitations
+
+Simple HTTP mode maps one HTTP request to one Varlink call. It therefore
+provides no way to issue `oneway` calls, to upgrade the connection with the
+`upgrade` flag, to pipeline multiple calls on one Varlink connection, or to
```
Skyb0rg007: Protocol upgrade is part of [HTTP semantics](https://datatracker.ietf.org/doc/html/rfc9110); I don't see why it couldn't be supported.

specs/varlink.md
Skyb0rg007: Perhaps reference [RFC 9421](https://www.rfc-editor.org/info/rfc9421/) as the mechanism for signing HTTP messages with ssh keys? This was only added to curl very recently though.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+  that interface as declared in the interface definition retrieved via
+  `org.varlink.service.GetInterfaceDescription`.
+
+* `GET /openapi/{socket}/{interface}` returns an
+  [OpenAPI 3.1](https://spec.openapis.org/oas/v3.1.0) document (JSON)
+  describing the `/call/{socket}/{method}` endpoints of that interface, with
+  schemas derived from the interface definition and documentation comments
+  carried over as descriptions.
+
+### WebSocket Mode
```
Skyb0rg007, line 1290: I don't think WebSocket mode is a good idea to formalize. WebSockets are hard to proxy, WAFs don't support it, and its implemented differently across HTTP versions. If systemd wants to support it, then fine, but standardizing it seems like a bad idea.

specs/varlink.md
```diff
@@ -0,0 +1,1348 @@
...
+
+In WebSocket mode the client performs an HTTP upgrade to the WebSocket
+protocol ([RFC 6455](https://www.rfc-editor.org/rfc/rfc6455)) on
```
Skyb0rg007: If you're allowing HTTP/2 or HTTP/3, you need to cite additional RFCs: 6455 is HTTP/1.1 specific.

Skyb0rg007 (commented):
Some comments

specs/varlink.md
mvo5: The `varlink-http-bridge`  is currently using https://github.com/oxidecomputer/sshauth for this (which is not using this) - adding/switching to https://www.rfc-editor.org/info/rfc9421/ would be super interesting.

specs/varlink.md
mvo5: One driver for this is the protocol-upgrade support in varlink. Everything else can be done nicely with standard http but a protocol upgrade needs a bi-directional stream so websockets seems to be the cleanest answer (there are other ways like a custom "Upgrade: varlink" or doing long lived h2 streams but those have their own deployment problems). Maybe this could state that for all non-protocol-upgrade "plain" http is sufficient?

specs/varlink.md
poettering: I tried to figure out based on that github issue, and I still don't get it. Please explain.

specs/varlink.md
poettering: dunno, the IDL quite easy to parse. Have two sources of truth sounds, and two languages sounds worse to me, compared to having a parser for the simple IDL.

specs/varlink.md
mvo5: I like the suggestion! I opened https://github.com/systemd/varlink-http-bridge/pull/117 with one possible way to implement this: client adds `?oneway=true` and the bridge sets the oneway flag and replies 204. I was thinking about a custom header too instead of ?oneway=true but it seems there is no standard header and a custom one seems to add nothing over the query parameter. But no strong opinion here, ideas welcome :)

specs/varlink.md
Skyb0rg007: I would have imagined translating Varlink protocol upgrade into HTTP Upgrade (to WebSockets); that way you avoid the downsides of WS whenever the server doesn’t support it.

specs/varlink.md
mvo5: I looked a bit into this and I think adding the MUST here would be beneficial:
* systemd itself is fixed https://github.com/systemd/systemd/pull/38922
* slightly more consistent: earlier we say empty objects and null are not equivalent but here we add an exception (and in l591,l614 too)
* slightly easier to generate typed client code as it no longer needs to deal with the nil vs empty-value case (e.g. the varlink-http-bridge would also benefit as it has a single clear rule how to encode now)

poettering:
I force pushed a new version with a number of changes. I addressed all review threads I marked as resolved. I also changed the "sockets" strings in the HTTP urls to "services", after discussing with @mvo5. PTAL.

I did not address any requests for protocol additions. That should probably be prepped as spec update once this version is merged.

specs/varlink.md
Skyb0rg007: Does varlink-http-bridge currently sign the Accept: header? That controls the `more` parameter, so it’s important for integrity.

A standard should at least list explicitly which HTTP headers and pseudo-headers to sign (see RFC 9421 for details), if not for RFC 9421 support then for accurately describing the important parts of the HTTP encoded message.

specs/varlink.md
poettering: > * slightly more consistent: earlier we say empty objects and null are not equivalent but here we add an exception (and in l591,l614 too)

I think this is a misunderstanding: while it is true that they are not generally equivalent, we also say that absence of a field shall mean "default" or "does not apply". Or in other words: null/absence for a specific field *is* to be understood equivalent to its default value. For `parameters` the default is simply `{}`. hence absence, null or `{}` are equivalent for `parameters` – not for other fields, but for this one it is.

The spec actually uses the IDL to describe the call/reply definition itself:

https://github.com/poettering/specifications2/blob/d3ee3f243524612fdd3afd43e3ccd4888d2ede80/specs/varlink.md#call

if we actually wanted to do what @zeenix wants, then we simply couldn't express this anymore, because there's no way in the IDL to say that `{}` is not allowed for a field of type `?object`.

or are you suggesting we add that? yucketyyuck

specs/varlink.md
```diff
@@ -0,0 +1,1395 @@
...
+    -H "Content-Type: application/json" -d '{}'
+{"Hostname":"myhost","StaticHostname":"myhost",…}
+```
+
+(The base URL is `https://host:1031/waldo/` in this example.)
+
+#### Streaming Replies
+
+Methods that reply multiple times (i.e. use the `more` flag) are invoked in
+simple mode by sending the request header `Accept: application/json-seq`. The
```
Skyb0rg007, line 1237: We should explicitly say that the valid `Accept` values are `Accept: application/json-seq` for `more=true`, and `Accept: application/json` (the default) for `more=false`; all other values are protocol errors.

specs/varlink.md
poettering: i'd really try to understand why you are bothered by the "nul vs empty-value" case. can't you just say in your code too in a first step that a parameters field that is absend or null is mangled into `{}`, and then from that point on everything is uniform?

specs/varlink.md
```diff
@@ -0,0 +1,1395 @@
...
+Each successful reply is encoded as one record containing its output
+parameters (unwrapped, as above). The `continues` flag is not transmitted: the
+sequence ends after the final reply, i.e. when the HTTP response body ends.
+An error reply terminates the sequence with a record of the form
+`{"error": "<fully-qualified error name>", "parameters": {…}}`; a transport
+failure during streaming terminates it with a record of the form
+`{"error": "<message>"}`. Since the HTTP status line has already been sent
+when such an error occurs, the status is `200 OK` regardless.
+
+The interface definition carries no indication whether a method supports or
```
Skyb0rg007:
```suggestion
The interface definition at this time carries no indication whether a method supports or
```

Because this will almost certainly be standardized in a later version of this spec.

specs/varlink.md
```diff
@@ -0,0 +1,1395 @@
...
+    -d '{"service":"io.systemd.Multiplexer"}' | jq --seq
+```
+
+#### Limitations
+
+Simple HTTP mode maps one HTTP request to one Varlink call. It therefore
+provides no way to issue `oneway` calls, to upgrade the connection with the
+`upgrade` flag, to pipeline multiple calls on one Varlink connection, or to
+pass file descriptors. Clients that need these features should use WebSocket
+mode.
```
Skyb0rg007: Phrased poorly: WebSocket mode doesn't support passing file descriptors either.

specs/varlink.md
```diff
@@ -0,0 +1,1395 @@
...
+sequence ends after the final reply, i.e. when the HTTP response body ends.
+An error reply terminates the sequence with a record of the form
+`{"error": "<fully-qualified error name>", "parameters": {…}}`; a transport
+failure during streaming terminates it with a record of the form
+`{"error": "<message>"}`. Since the HTTP status line has already been sent
+when such an error occurs, the status is `200 OK` regardless.
+
+The interface definition carries no indication whether a method supports or
+requires `more`, hence the bridge cannot determine this up front. If a method
+that requires `more` is invoked without the `application/json-seq` accept
```
Skyb0rg007: Does this require `Accept: application/json-seq`, or just that the MIME type is `application/json-seq` (ex. is `Accept: application/json-seq;charset=UTF-8` acceptable?)

Skyb0rg007 (commented):
Some other comments on the HTTP spec, where there's ambiguity due to the complexities of HTTP. I agree with the ssh auth removal; that definitely needs more work before it becomes a UAPI spec.

specs/varlink.md
Skyb0rg007: I'll continue this discussion here: https://github.com/systemd/varlink-http-bridge/issues/119

specs/varlink.md
```diff
@@ -0,0 +1,1395 @@
...
+
+* **Transport failure.** If the bridge cannot connect to the service's
+  socket, or the connection fails while the call is in progress,
+  status `502 Bad Gateway` is returned. Malformed requests (invalid
+  JSON body, unknown service) yield `400 Bad Request` or `404 Not
+  Found` as appropriate. Requests that fail authentication yield `401
+  Unauthorized`.
+
+  In these cases, the body is a JSON object of the form
+  `{"error": "<human readable message>"}`.
```
Skyb0rg007: Using the same key for the machine-readable and the human-readable error is a bad choice. Perhaps `{"errorString": "<human readable message>"}`?

Skyb0rg007: As a concrete reason why: as mentioned by @mrc0mmand, having a dedicated "ProtocolError" would be great and could be a future addition to the spec. But as currently written this would be a breaking change since the human readable error would be replaced with a machine readable one (not allowing a human-readable fallback).

specs/varlink.md
chergert: > field that is absent or null is mangled into {}

This is exactly what I did in varlink-glib, FWIW.

specs/varlink.md
aphedges: Thanks for handling this!

specs/varlink.md
Skyb0rg007: Now that I think about it, the `oneway` semantics seem to be equally well supported by HTTP as it is for WebSockets: both protocols require explicit acknowledgement by the remote device before the `oneway` message can be discarded, its just at a different protocol layer (L4 with the TCP message queue vs L7 via HTTP worker thread). Sending `oneway` and then a non-`oneway` message still requires the server to read them in order and acknowledge both.

specs/varlink.md
poettering: according to specs, neither of these mime types even defines charset=. So i guess we should recommend against it, but of course if people send it a conformant http server should be able to process it and ignore it.

specs/varlink.md
poettering: i changed the field to `message` now.

poettering:
I force pushed a new version. Changes:

it dictates the Accept: http header to send explicitly now.

the "error" field when human readable is changed to "message".

references to fd passing via http are removed.

ptal

brauner (approved):
We're mostly just moving an existing spec over under the neutral uapi umbrella. I think we're fine to merge this. Any follow-up fixes and additional should go in via separate pull requests. So I intend to merge this.

poettering:
Merged.

So, the various spec additions suggested here, can you please file individual PRs for those, so they aren't forgotten?

I think if you have a proper spec change you'd like to propose, just file it as PR here against the spec. (or if need be an issue, if you just want to point out a bug in the spec)

If you have more general ideas without any specific change in mind, i think discussing that over on the varlink repo makes more sense. already because a bunch of things are already open there.

specs/varlink.md
zeenix: > I think this is a misunderstanding: while it is true that they are not generally equivalent, we also say that absence of a field shall mean "default" or "does not apply".

So essentially, you're saying they're not equivalent really but they are becuase you say so? :thinking: 

> if we actually wanted to do what @zeenix wants, then we simply couldn't express this anymore, because there's no way in the IDL to say that `{}` is not allowed for a field of type `?object`.

I don't see why IDL needs to say that at all, if the spec specifies a very specific encoding of `?object`. i-e the spec defines the meaning of null.

> I tried to figure out based on that github issue

Did you look into the serde enum representations docs?

---

In any case, I will try to get the serde issue fixed but I can totally understand if the maintainer isn't interested. One big reason would be that not everyone would want `{}`, null and missing field to be equivalent. E.g imagine a JSON-based IPC protocol where the spec writers made it very specific (like it really should be IMO). Now they'd want deserialization to end in error if `{}` was provided if they had chosen one of the other encoding.

zeenix: > > field that is absent or null is mangled into {}
> 
> This is exactly what I did in varlink-glib, FWIW.

I'd probably have done exactly the same if I were to use my own JSON deserializer. Perhaps I should do that anyway but it's too much work and adding to maintainance burdon for this one very specific thing that can easily be avoided if people would listen to me and accept that this makes sense in general and not only for serde.

poettering: > So essentially, you're saying they're not equivalent really but they are becuase you say so? 🤔

I am saying that for *this* field we deault to `{}` if unset or nullified.

varlink IDL does this for all nullable fields: let's say you have a method that takes a field `timeoutSec` of type `?int`. You make it nullable, because while some clients might want to configure a timeout for whatever this does explicitly, most probably just want the default timeout. You then document, hey if you don't pass this, then default is 15s. Excellent. So for this field now suddenly absence, null and `15` is the same. 

The same thing applies to `parameters`: we document that the default parameter list is the empty one, and hence for this field, absence, null, and `{}` is the same.

That doesn't mean that universally the three are the same for all fields. it just means for the `parameters` fields, yes it is the same.

And hence you shouldn't really need anything from serde here: just make sure that your parser defaults to `{}` if `parameters` is absent/null. Similar to how `more`, `oneway` and `upgrade` all default to `false` if they are absent/null.

zeenix: > And hence you shouldn't really need anything from serde here: just make sure that your parser defaults to `{}` if `parameters` is absent/null.

I can do manual serde::Deserialize impl for enums in zlink itself (that's never been an issue, really) but most users would (for very good reason) want to use the derive macros. So to really get around this issue, there are only 2 options:

1. Have serde treat these 3 things the same way during deserialization. I just had Claude prepare patches for this and it works perfectly and I'll submit them soon after I've had time to understand the code and submit a PR for it. However, as I wrote this may not be accepted for the reason I mentioned above.
2. I write my own deserializer for JSON, which isn't ideal either.

> I am saying that for _this_ field we deault to `{}` if unset or nullified.

That's just a different way of saying that we treat all those 3 as the same.

specs/varlink.md
Skyb0rg007: Hmm, I don’t like the use of `MUST`, since many existing `curl` samples don’t include the `Accept` header. Is there any reason why “SHOULD + server treats missing as application/json” didn’t work? @poettering

Skyb0rg007: Actually the current doc even makes this mistake, including a `curl` example without explicit Accept header…

poettering:
> One thing I'd like to consider is when you're in a situation that you only have a single socket you can use, and what to multiplex over it (say USB device connection). In `varlink-glib` I ended up creating a simple muxer. Obviously most situations can just create a socket, so this is more about when that is not an option.
> 
> https://libgvarlink-1f981c.pages.gitlab.gnome.org/varlink-glib-1/protocol-muxer.html

So i thought about this for a bit, and put some notes/ideas up, see this: 

https://github.com/uapi-group/specifications/issues/232

in some ways it's quite close to @chergert's multiplexer.

poettering:
> If you have more general ideas without any specific change in mind, i think discussing that over on the varlink repo makes more sense. already because a bunch of things are already open there.

i take that back. the varlink repo is kinda for the varlink *website* by everything how it is set up. I think spec stuff should be discussed here entirely hence, because this one is really about spec'ing out stuff. Hence, please, when you have spec extension suggestions, let's do it here.

PR #229, opened 2026-09-10
Fix curl example in Varlink spec

Skyb0rg007:
The requirement that Simple HTTP requests include the `Accept: application/json` header was a last minute change, and the included curl example was not updated.

This PR just fixes the example by having curl pass `Accept: application/json` (the `--json` flag sets both `Content-Type` and `Accept`); the alternative fix is to change the spec from:

> Every request MUST be made with an `Accept: application/json` header, except if a streaming reply is expected, in which case `Accept: application/json-seq` should be sent.

to

> Every request SHOULD be made with an `Accept: application/json` header, except if a streaming reply is expected, in which case `Accept: application/json-seq` should be sent. An absent `Accept` header MUST be treated by an HTTP server as if it were set to `application/json`.


mvo5 (approved):
Thank you!

Skyb0rg007:
Before this is merged: @mvo5 do you think that the `SHOULD` + default to `application/json` is a better option? Looking at the https://github.com/systemd/varlink-http-bridge README, every example would need to change as `curl -s <url>` is now invalid.

poettering:
I think we probably should do both? the example line looks more correct to me either way

Skyb0rg007:
Hmm, looking into the HTTP `Accept` header shows that it’s a [content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Content_negotiation), not a requirement:

```http
Accept: application/json, application/json-seq
```

is completely valid, and the server is supposed to choose whichever they support. If the Simple HTTP spec wants to adhere to HTTP Semantics it seems like we can have servers allow this and just respond with their preferred response format, though it would complicate HTTP->Socket bridging since the bridge doesn’t know whether to send `"more": true` or not.

It’s probably more “HTTP-native” to add something like a `?more=true` query parameter.

Edit: I didn't remember that Varlink's `more` argument can be legally ignored: in that case it does fit the "negotiation" aspect. I'll update this PR.

Skyb0rg007:
I'm still not exactly happy with the mapping of Varlink to HTTP, since `Accept` is part of HTTP's "server-driven content negotiation" feature, which isn't really how Varlink works.

For example, say a client and server implement a CBOR extension that lets clients + servers send their Varlink data encoded using CBOR instead of JSON. This could be negotiated by having the client send `Accept: application/json, application/cbor`; a normal server will just return JSON but a server that supports the extension could return CBOR.

The use here is different because the data returned when `application/json-seq` is allowed to be different. It's client-driven, so it shouldn't be using any of the headers specifically setup to be used for **server-driven** content negotiation (`Accept`, `Accept-Language`, `Accept-Encoding`). Since the client is the one that chooses whether `more` is set to true/false, we should be using something like a query parameter.

mvo5:
Hey, sorry for my slow reply (had a busy few days). I looked at this a bit and I agree, I think adding `?more=true` (and probably plain `?more`) is probably the right call.  I am happy to update the bridge for this.

Skyb0rg007:
> Hey, sorry for my slow reply (had a busy few days).

No worries! AFAIK you own the only varlink http bridge implementation so its not like a change in the standard will cause any integration problems.

> adding `?more=true` (and probably plain `?more`)

We should definitely require clients to use `?more=true`; query params aren't part of an IETF standard but `application/x-www-form-urlencoded` ([W3C](https://www.w3.org/TR/REC-html40/interact/forms.html#h-17.13.4.1)) does require equal signs.

mvo5:
[..]
> > adding `?more=true` (and probably plain `?more`)
> 
> We should definitely require clients to use `?more=true`; query params aren't part of an IETF standard but `application/x-www-form-urlencoded` ([W3C](https://www.w3.org/TR/REC-html40/interact/forms.html#h-17.13.4.1)) does require equal signs.

Thanks! I opened https://github.com/systemd/varlink-http-bridge/pull/134 with a quick fix for this to not use the accept header anymore for more (and some of the --json curl magic :)

keszybz (approved):
LGTM too.

Skyb0rg007:
> LGTM too.

Ahh! Was about to change this to the new `?more=true` variant. I'll make another PR.

PR #233, opened 2026-09-15
varlink: mention the /run/varlink/registry/ directory in the spec

mvo5:
Systemd implements the /run/varlink/registry/ directory in [1]. It was not mentioned in the spec but it should be listed here as it complements the user.varlink xattr.

[1] https://github.com/systemd/systemd/pull/40590


keszybz (approved):
Looks all good.

